Issue #39, opened 2026-02-12
Optimize skill index query with LATERAL join


lfiaschi:
## Summary

Replace the `ROW_NUMBER()` window function in `fetch_all_skills_for_index` with a `LATERAL` subquery join for better query performance.

## Problem

The current query computes `ROW_NUMBER() OVER (PARTITION BY skill_id ORDER BY semver_major DESC, ...)` across **all** version rows, then filters to `rn = 1`. This scans and ranks every version row even though we only need the top-1 per skill.

## Proposed Solution

Use a `LATERAL` subquery that, for each skill, performs a single index-range scan on the existing `idx_versions_skill_semver_parts` composite index and stops after 1 row. Cost is O(skills) instead of O(versions).

## Spec

See `specs/pr-31-lateral-join-optimization.md`

## Scope

- Single-file change: `server/src/decision_hub/infra/database.py`
- No migrations, no API changes, no CLI changes
- Function signature and return shape are identical

PR #191, opened 2026-02-24
fix: patch Cisco scanner for Gemini + detect silent LLM degradation

### LINKED ISSUE #187: skill-scanner LLM analyzer silently fails with Gemini
## Problem

The `cisco-ai-skill-scanner` LLM analyzer and meta-analyzer silently fail when using Google Gemini (`gemini-2.0-flash`) as the LLM provider. The scan completes and returns a grade, but it's based only on static/behavioral rules — no LLM judgment runs.

Introduced in #146.

## Root cause

The Google GenAI SDK's structured output API rejects the scanner's JSON response schema because it uses union types (`['string', 'null']`) which aren't valid for Gemini's `Schema` type enum. Gemini expects a single type like `'STRING'`, not an array.

Error (repeated for 4 schema fields):
```
LLM analysis failed: 4 validation errors for Schema
properties.findings.items.properties.aisubtech.type
  Input should be 'TYPE_UNSPECIFIED', 'STRING', 'NUMBER', 'INTEGER', 'BOOLEAN', 'ARRAY', 'OBJECT' or 'NULL'
  input_value=['string', 'null'], input_type=list
```

The scanner's `ProviderConfig` detects Gemini and routes to the Google GenAI SDK (`use_google_sdk=True`), but the schema passed to `client.models.generate_content()` was designed for OpenAI/Anthropic and is incompatible with Gemini's structured output format.

## Impact

- All three publish paths (CLI upload, crawler, tracker) call the LLM analyzer
- The LLM + meta-analyzer failures are caught and swallowed — the scan continues with static-only results
- This means **Scenario C (full pipeline)** effectively degrades to **static-only** without any warning beyond a stderr print
- The skill-scanner README only lists Anthropic and OpenAI as supported `--llm-provider` values; Gemini support appears to be undocumented/experimental

## Reproduction

```python
from decision_hub.settings import create_settings
from decision_hub.domain.skill_scanner_bridge import scan_skill_dir
from pathlib import Path

settings = create_settings()  # uses GOOGLE_API_KEY + gemini-2.0-flash
result = scan_skill_dir(Path('bootstrap-skills/dhub-cli'), settings)

# result.analyzers_used includes 'llm_analyzer' and 'meta_analyzer'
# but result.meta_analysis is None — both failed silently
```

## Options

1. **Switch to Anthropic** (`claude-sonnet-4-6`) — first-class supported provider. Requires adding `ANTHROPIC_API_KEY` to server settings and Modal secrets for the scanner (separate from the eval pipeline key).
2. **Switch to OpenAI** (`gpt-4o`) — first-class supported provider. Requires adding an OpenAI API key.
3. **File upstream bug** against `cisco-ai-defense/skill-scanner` for Gemini schema incompatibility and run static-only until fixed.
4. **Run static-only intentionally** — remove `LLMAnalyzer` and `MetaAnalyzer` from `_build_analyzers()` until a supported provider is wired in.

## Recommendation

Option 4 (static-only) as an immediate fix to avoid silent failures, then Option 1 or 3 as the proper solution.

---

### Grade "B" is unreachable

`skill_scanner_bridge.py:50-57` — The severity-to-grade mapping only produces A, C, and F:

| Severity | Grade |
|----------|-------|
| CRITICAL, HIGH | F |
| MEDIUM | C |
| LOW, INFO, SAFE | A |

Grade "B" is defined in `SafetyGrade` but no scanner severity maps to it. Intentional simplification or oversight?

### Near-duplicate quarantine functions (DRY violation)

`registry_service.py` has two quarantine functions that are ~90% identical:
- `quarantine_rejected_scan()` (line 192) — raises `HTTPException(422)` at the end
- `quarantine_scan_rejection()` (line 244) — does not raise

The names are also confusingly similar. These should share a private helper, with the HTTP-raising variant calling it and then raising.

### Overly defensive `getattr` on typed Settings

`skill_scanner_bridge.py:80-81`:
```python
api_key = getattr(settings, "google_api_key", None)
model = getattr(settings, "gemini_model", "gemini-2.0-flash")
```

`settings` is always a `Settings` instance with these fields present. The `Any` type hint on the parameter hides this. Should use `settings: Settings` and plain attribute access.

### Private function imported across modules

`registry_routes.py:23` imports `_scan_result_to_audit_fields` — a private function by Python convention. If it's used across modules, drop the leading underscore.

cc @maresb

maresb:
## Summary

Re-applies the skill-scanner integration (#146, #185) that was reverted in #188, with two fixes for the Gemini schema incompatibility and silent LLM failure issues.

- **Gemini schema fix**: Monkey-patch `LLMRequestHandler._sanitize_schema_for_google` to convert JSON Schema union types (`["string", "null"]` → `{"type": "STRING", "nullable": true}`) that the Google GenAI SDK rejects. Upstream fix: [cisco-ai-defense/skill-scanner#39](https://github.com/cisco-ai-defense/skill-scanner/pull/39).
- **Silent degradation detection**: `_check_llm_degradation()` detects when LLM analyzers were configured but produced no output (no meta-analysis, no LLM findings), and injects a visible `LLM_DEGRADED` INFO finding. Works around the upstream silent-failure pattern ([cisco-ai-defense/skill-scanner#38](https://github.com/cisco-ai-defense/skill-scanner/issues/38)).
- **Version pin**: `cisco-ai-skill-scanner==1.0.2` pinned with inline removal instructions for when the upstream fix is released.

## Remaining cleanup (follow-up PR)

- Near-duplicate quarantine functions (DRY violation)
- `getattr` on typed `Settings` in `_build_analyzers`
- Private `_scan_result_to_audit_fields` imported across modules

## Test plan

- [x] 10 new tests for `_fix_gemini_union_types` (union conversion, nesting, all JSON types, real scanner schema)
- [x] 7 new tests for `_check_llm_degradation` (no-op cases, detection, finding injection, field preservation)
- [x] 649 server tests pass
- [x] 227 client tests pass

Closes #187 (Gemini schema incompatibility + silent degradation)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/server/src/decision_hub/infra/database.py
+++ b/server/src/decision_hub/infra/database.py
@@ -1696,12 +1696,12 @@ def _build_skills_filters(
     latest_eval_status column on skills_table directly.
     """
     if search:
-        pattern = f"%{search}%"
+        pattern = f"%{_escape_like(search)}%"
         base = base.where(
             sa.or_(
-                skills_table.c.name.ilike(pattern),
-                skills_table.c.description.ilike(pattern),
-                organizations_table.c.slug.ilike(pattern),
+                skills_table.c.name.ilike(pattern, escape="\\"),
+                skills_table.c.description.ilike(pattern, escape="\\"),
+                organizations_table.c.slug.ilike(pattern, escape="\\"),
             )
         )
     if org_slug:
@@ -2055,7 +2055,7 @@ def fetch_org_stats(
 
     # Filter on non-aggregate columns before grouping (WHERE, not HAVING)
     if search:
-        stmt = stmt.where(organizations_table.c.slug.ilike(f"%{search}%"))
+        stmt = stmt.where(organizations_table.c.slug.ilike(f"%{_escape_like(search)}%", escape="\\"))
 
     if type_filter == "orgs":
         stmt = stmt.where(organizations_table.c.is_personal == sa.false())


lfiaschi:
Closing this PR — it's now 16+ days stale with merge conflicts and 40+ PRs merged to main since it was opened. The gauntlet pipeline has received significant improvements in that time (56% false positive reduction, security hardening, credential detection fixes), making much of this diff outdated.

The valuable research here (40-run LLM benchmarking study, Gemini union-type fix, bridge architecture) should inform a fresh approach. Recommendation: rethink the scanner integration against the current state of main, and adopt incrementally — contribute upstream fixes, add scanner as a parallel signal behind a feature flag, collect data, then migrate.

Closing in favor of a future incremental approach.

PR #232, opened 2026-03-02
Add structured JSON logging and enhanced request context

lfiaschi:
## What changed
Enhanced the logging system to support structured JSON output and enriched request context with user, org, and skill information extracted from JWT tokens and URL paths. Added a `/health` endpoint for service and database connectivity checks.

## Why
Closes #(issue number)

Structured JSON logging is essential for production observability—it enables efficient log aggregation, filtering, and analysis in centralized logging systems. The enhanced request context (user, org_slug, skill_name, response_size) improves correlation and debugging by automatically capturing relevant metadata without requiring explicit logging at each call site. The health endpoint provides standard operational visibility for load balancers and monitoring systems.

## How to test
1. Run `make test` to verify all unit tests pass, including new tests for:
   - JSON and text logging format configuration
   - JWT username extraction from Bearer tokens
   - URL path regex for org/skill extraction
   - Health endpoint connectivity checks
2. Verify text format (default) still produces human-readable logs
3. Enable JSON format via `log_format="json"` setting and confirm valid JSON lines are written to stderr
4. Test `/health` endpoint returns 200 when database is reachable, 503 when unreachable

## Checklist
- [x] No breaking API changes (new optional parameter with sensible default)
- [x] Database migration included (if schema changed) — N/A

https://claude.ai/code/session_01GAzoFi3fc6zTqqqBMWvxRK

### DIFF
--- a/server/src/decision_hub/logging.py
+++ b/server/src/decision_hub/logging.py
@@ -11,6 +11,7 @@ import logging
 import re
 import sys
 import time
+import traceback
 import uuid
 
 from loguru import logger
@@ -72,10 +73,11 @@ def _json_sink(message) -> None:
     }
     # Include bound context fields (request_id, user, org_slug, etc.)
     for key in ("request_id", "user", "org_slug", "skill_name", "response_size"):
-        if extra.get(key):
+        if key in extra:
             entry[key] = extra[key]
     if record["exception"]:
-        entry["exception"] = str(record["exception"])
+        exc_type, exc_value, exc_tb = record["exception"]
+        entry["exception"] = "".join(traceback.format_exception(exc_type, exc_value, exc_tb))
     sys.stderr.write(json.dumps(entry, default=str) + "\n")
     sys.stderr.flush()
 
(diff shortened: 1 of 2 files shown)


lfiaschi:
Addressed all four review findings in fc505a8:

1. **response_size never in logger context**: Fixed by using `logger.bind(response_size=response_size)` on the summary log line so it appears as a structured JSON field.
2. **Truthiness check drops zero-valued fields**: Changed `extra.get(key)` to `key in extra` so falsy values like `0` are preserved.
3. **Regex falsely matches non-skill org paths**: Split the single regex into two explicit alternatives (`/v1/skills/{org}/{skill}` and `/v1/orgs/{org}/skills/{skill}`) so paths like `/v1/orgs/acme/profile` no longer match.
4. **JSON sink loses exception traceback**: Replaced `str(record["exception"])` with `traceback.format_exception()` for a proper formatted traceback string.

Added tests for zero-valued fields, exception formatting, and non-skill org paths.

Issue #332, opened 2026-03-26
META 327-331: Cisco AI Skill Scanner integration and Gauntlet replacement


maresb:
## Overview

Integrate the [Cisco AI Skill Scanner](https://github.com/cisco-ai-defense/skill-scanner) as the safety analysis system for Decision Hub, replacing the existing Gauntlet. The full scope is broken into incremental, manageable tasks.

## Roadmap

```mermaid
graph LR
    I1["#327: Merge scanner\ninfrastructure"] --> I2["#328: Isolate\nGauntlet"]
    I1 --> I3["#329: Fresh batch\n+ stats"]
    I3 --> I4["#330: Fix pymc-labs\nskills"]
    I2 --> I5["#331: Remove Gauntlet\n+ delete flags"]
    I3 --> I5
    I4 --> I5
```

## Issues

1. **#327 — Merge Cisco scanner infrastructure (PR #307)**: Add scanner DB tables, bridge, scan-on-publish, backfill script, API endpoint, and frontend components. Scanner disabled behind feature flags.
2. **#328 — Isolate the Gauntlet for replacement**: Add `enable_gauntlet` flag, modularize all touchpoints, wire Cisco data as the alternative path.
3. **#329 — Run fresh scan batch and compute distribution stats**: Enable scanner in dev, run ~200 skills, compute verdict/risk/severity distributions for validation.
4. **#330 — Ensure `pymc-labs`-org skills pass scanner validation**: Fix valid findings in `pymc-labs` skills, document any false positives.
5. **#331 — Remove Gauntlet and delete all feature flags**: Final cleanup — delete Gauntlet code, remove all flags, update docs.

## Key decisions

- The Cisco scanner **replaces** the Gauntlet — it does not run alongside it permanently
- Cisco results are **not remapped** to A-F letter grades — the retrieval system and UI will be adapted to use the scanner's native verdict/risk semantics
- Feature flags allow incremental rollout: scanner can be enabled/disabled independently of Gauntlet

---

Addendum todos:
- [ ] address #333
- [ ] close https://github.com/pymc-labs/decision-hub/pull/334 (without merging)

PR #307, opened 2026-03-14
feat: integrate Cisco AI Skill Scanner infrastructure

maresb:
## Summary

Integrates the [Cisco AI Skill Scanner](https://github.com/cisco-ai-defense/skill-scanner) into Decision Hub, adding the full infrastructure for scanning skills on publish and displaying results in the UI. The scanner is **disabled by default** behind feature flags — enable when ready for validation.

Closes #327
Part of #332

### Backend
- **Database**: `scan_reports` and `scan_findings` tables ([migration](server/migrations/20260313_173012_create_scan_tables.sql))
- **Domain models**: `ScanReport` and `ScanFinding` frozen dataclasses
- **Scanner bridge**: [`skill_scanner_bridge.py`](server/src/decision_hub/domain/skill_scanner_bridge.py) — wraps the Cisco scanner SDK, maps results to domain models, stores in DB
- **Publish pipeline**: Scan-on-publish in [`execute_publish()`](server/src/decision_hub/domain/publish_pipeline.py), gated by `enable_cisco_scanner` setting (default: `False`)
- **Backfill script**: [`backfill_scan_reports.py`](server/src/decision_hub/scripts/backfill_scan_reports.py) with `--workers` and `--limit` flags, plus `make backfill-scan-reports` target
- **API endpoint**: `GET /v1/skills/{org}/{skill}/scan-report` — returns latest scan report with findings
- **Settings**: `enable_cisco_scanner` (bool, default `False`), `cisco_scanner_policy` (strict/balanced/permissive)

### Frontend
- **Scanner report component**: [`ScannerReport`](frontend/src/components/ScannerReport.tsx) integrated into the skill detail page audit tab and sidebar
- **Feature flag**: `SHOW_SCANNER_REPORT` in [`featureFlags.ts`](frontend/src/featureFlags.ts) (default: `false`)
- **Stale-version warning**: When a scan report is from an older skill version, the UI displays a warning

### Tests
- Unit tests for the scanner bridge ([`test_skill_scanner_bridge.py`](server/tests/test_domain/test_skill_scanner_bridge.py))

## Feature flags

Both flags default to **off** so this can be merged without affecting existing behavior:

| Flag | Location | Default | Controls |
|------|----------|---------|----------|
| `enable_cisco_scanner` | `server/src/decision_hub/settings.py` | `False` | Whether scans run during publish |
| `SHOW_SCANNER_REPORT` | `frontend/src/featureFlags.ts` | `false` | Whether scan results display in the UI |

## Architecture

[...]

PR #334, opened 2026-03-30
feat: add arXiv benchmark script for Cisco scanner validation

maresb:
## Summary

- Adds `server/scripts/run_arxiv_benchmark.py` — materializes each test case from the arXiv 2602.06547 malicious skill test set as a temp directory and scans via the full LLM + MetaAnalyzer pipeline
- Split out from #307 per review feedback to keep the infrastructure PR focused

Part of #332 (Cisco scanner rollout).

Made with [Cursor](https://cursor.com)


maresb:
I think that we will ever merge this. This was primarily intended for side-by-side comparison between Gauntlet and Cisco. Keeping this around for reference until #332 is complete.

PR #339, opened 2026-04-18
refactor: DRY up row-to-model mappers in database.py

lfiaschi:
## Summary

This PR came out of a codebase review (architect / principal-engineer pass) focused on finding high-leverage, low-risk cleanups in the server.

`server/src/decision_hub/infra/database.py` had **12 near-identical `_row_to_<Model>` mappers** — each one hand-unpacking every dataclass field with `row.<name>`. Every time someone added a column they had to edit two places (the frozen dataclass _and_ the mapper), and drift between the two was silent.

The PR collapses all 12 into thin one-line wrappers around a single generic helper:

```python
def _row_to_model(row: sa.Row, model_cls: type[_T]) -> _T:
    mapping = row._mapping
    field_names = {f.name for f in dataclasses.fields(model_cls)}
    return model_cls(**{k: mapping[k] for k in field_names if k in mapping})
```

The helper populates any dataclass field whose name matches a column in `row._mapping`. Extra columns (e.g. `fts_rank`, `vec_dist` on search queries) are ignored; absent fields fall back to the dataclass default (important for `User.github_orgs`, which is populated from JWT claims and not the DB).

### Latent bug fixed as a side-effect

`_row_to_skill_tracker` never assigned `consecutive_permanent_failures`, so any caller reading that field off a DB-loaded `SkillTracker` would see the dataclass default of `0` instead of the real value persisted in `skill_trackers.consecutive_permanent_failures`. The generic helper fixes this automatically, and a regression test guards against it.

### Small adjacent cleanup

Also extracts the triplet `ORDER BY semver_major DESC, semver_minor DESC, semver_patch DESC` into a module-level `_SEMVER_DESC` tuple. Used in three places (`_refresh_skill_latest_version`, `resolve_version`, `resolve_latest_version`).

## Numbers

- `database.py`: **3465 → 3337 lines** (−128 LOC)
- Diff: **+435 / −211** overall
- Adds 24 targeted unit tests, including a `@parametrize` drift guard over 7 representative models
- `_row_to_scan_report` and `_row_to_scan_finding` are left untouched — they unpack nested JSON fields and don't fit the generic shape

## Out of scope (flagged for follow-up)

The review surfaced several larger items I deliberately did not bundle here:

- `database.py` (still 3337 lines / 105 functions) could reasonably split into `tables.py` / `skills.py` / `trackers.py` / `evals.py`.
- CLI `registry.py` is 1912 lines and mixes publish / install / search / management.
- `gauntlet.py` (1355 lines) could become a `gauntlet/` package keyed by check type.
- `tracker_service.check_all_due_trackers()` is 301 lines and `publish_pipeline.execute_publish()` is 264 — both benefit from extraction but are high-stakes.
- Untested modules: `seo_routes`, `taxonomy_routes`, `repo_utils`, `crawler/*`.

Happy to tackle any of those in separate, smaller PRs.

## Test plan

- [x] `uvx ruff check` / `ruff format --check` — clean
- [x] `pytest server/tests/test_infra server/tests/test_api server/tests/test_domain server/tests/test_scripts -m "not slow"` — **951 passed, 34 deselected**
- [x] `pytest shared/tests` — **98 passed**
- [x] `pytest client/tests` — **291 passed, 29 skipped**
- [x] Existing `test_tracker_db.py::TestRowToSkillTracker` / `TestRowToTrackerMetrics` updated to populate `row._mapping` alongside attribute access — all 32 tracker tests still pass
- [x] New `test_row_to_model.py` asserts: exact-field population, extra-column-ignored, default-fallback-on-missing, TypeError-on-missing-required, and the `consecutive_permanent_failures` regression

## Behavioural change

None at any call site: every `_row_to_<X>(row)` function still exists and still returns the same model type. The only observable difference is that `SkillTracker.consecutive_permanent_failures` is now populated from the row instead of always being `0`.

https://claude.ai/code/session_015P8XHs55PZUwLxeKjk3Gdc


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation. This automated review-sweep PR overlaps heavily with the retained merge queue (#449, #448, #447, #446, #405, then #438, #443, #375). Unique fixes not covered by the retained set are catalogued for a follow-up best-of PR in #451. The branch is preserved, so nothing is lost.

PR #345, opened 2026-04-21
refactor(api): consolidate rate-limit enforcers; bound publish metadata

lfiaschi:
## Summary

Collapses **ten copy-pasted `_enforce_*_rate_limit` helpers** across `registry_routes.py`, `search_routes.py`, and `auth_routes.py` into a single `rate_limit(name)` dependency factory, consolidates **two duplicated `_parse_uuid` helpers** into `api/deps.py`, and adds a **`max_length=4096` cap on the publish `metadata` Form field** so a hostile client can't burn memory on multi-MB JSON parsing or wedge the log pipeline.

Net change: **-126 lines of duplication**, +40 lines of factory + tests, **+8 unit tests**, all 943 non-slow server tests still pass.

---

## Scope of the review that produced this PR

I did a principal-engineer deep dive across the whole monorepo — server API, domain, client CLI, shared, frontend, tests/observability. The full assessment surfaced many findings; this PR ships the **highest-leverage, low-risk cluster** that fits in a focused review. The rest is enumerated below under "Findings deferred to follow-up PRs" so nothing is lost.

## High-level codebase health

- **Architecture is sound.** Clean `client` / `server` / `shared` boundary with `dhub_core` as the single source of truth for manifest parsing and validation. Domain layer is mostly pure; infra is a thin adapter set.
- **Observability is good.** `loguru` with `{}` placeholders is respected (zero f-strings in logger calls), per-request 8-char request-id middleware, exception attachment via `.opt(exception=True)`. Missing pieces: Sentry/error reporting, OpenTelemetry, Prometheus metrics — observability gaps are at the edges, not the core.
- **Test pyramid is bottom-heavy (as intended).** 943 non-slow server tests + 311 client + 8 frontend. Strong factory pattern at `server/tests/factories.py`. Coverage enforced at 50%. Main gap: frontend tests aren't wired into CI.
- **The drag.** Hot files have drifted into god-modules (`infra/database.py` 3,465 lines, `domain/gauntlet.py` 1,355 lines, `cli/registry.py` 1,912 lines). Rate-limit plumbing is copy-pasted. Some silent-failure paths in the publish pipeline. Five near-identical Gemini callback builders in `publish_pipeline.py`.

---

### 1. `rate_limit(name)` factory (`server/src/decision_hub/api/rate_limit.py`)

Before: ten identical ~10-line functions, one per endpoint:

```python
def _enforce_list_skills_rate_limit(request: Request) -> None:
    state = request.app.state
    if not hasattr(state, "_list_skills_rate_limiter"):
        state._list_skills_rate_limiter = RateLimiter(
            max_requests=settings.list_skills_rate_limit,
            window_seconds=settings.list_skills_rate_window,
        )
    state._list_skills_rate_limiter(request)
# ...nine more just like it...
```

After: one factory, used via `Depends(rate_limit("list_skills"))`:

```python
@cache
def rate_limit(name: str) -> Callable[[Request], None]:
    state_attr = f"_{name}_rate_limiter"
    limit_field = f"{name}_rate_limit"
    window_field = f"{name}_rate_window"

    def enforce(request: Request) -> None:
        state = request.app.state
        limiter: RateLimiter | None = getattr(state, state_attr, None)
        if limiter is None:
            settings = state.settings
            limiter = RateLimiter(
                max_requests=getattr(settings, limit_field),
                window_seconds=getattr(settings, window_field),
            )
            setattr(state, state_attr, limiter)
        limiter(request)

    enforce.__name__ = f"enforce_{name}_rate_limit"
    enforce.__qualname__ = enforce.__name__
    return enforce
```

Design notes:

- **`@cache` per name** so repeated `Depends(rate_limit("x"))` declarations share the same callable identity — important for FastAPI's dependency cache.
- **Lazy init on first request** preserves the existing semantics (limiter created from settings only when actually hit; tests can swap `test_settings` without app restarts).
- **`getattr` on settings** makes typos fail loud at first hit (`AttributeError`) instead of silently disabling the limit — a covered test case.
- **Descriptive `__name__`** so FastAPI docs and tracebacks still identify which limit fired.

### 2. `parse_uuid_param` in `api/deps.py`

The same 5-line UUID-parse-or-422 helper was defined in both `registry_routes.py` and `tracker_routes.py`. Moved to `deps.py` and imported by both; call sites updated (3 in each file). Error detail is now consistent across the API.

### 3. `max_length=4096` on publish metadata Form (`registry_routes.py`)

```python
def publish_skill(
    # Bound the metadata blob so a malicious client cannot burn memory on
    # JSON parsing or wedge the log pipeline with a multi-MB payload. The
    # real body is ~1 KB (org_slug, skill_name, version, visibility,
    # source_repo_url, manifest_path); 4 KB leaves comfortable headroom.
    metadata: str = Form(..., max_length=4096),
    ...
):
```

Real publish metadata reads 6 short fields totalling ~1 KB. 4 KB gives comfortable headroom and blocks abuse. Rejected payloads return **422 before JSON parsing**, so `json.loads(...)` never sees the blob.

### 4. Tests added

- `test_rate_limit.py` gains a `TestRateLimitFactory` class with 7 tests: identity caching per name, distinct callables per name, lazy settings read, limiter cached on `app.state`, enforcement round-trip, `AttributeError` on missing settings field, descriptive `__name__`.
- `test_deps.py` gains `TestParseUuidParam` with happy-path and a 422 assertion that checks the error mentions both the parameter name and the offending value.
- `test_registry_routes.py` gains `test_publish_rejects_oversized_metadata` covering the new Form bound.

---

## Findings deferred to follow-up PRs

Captured so they aren't lost. Each one was verified against actual source (file:line); none of them are blocking for this PR.

### Server — correctness & security

1. **Publish rollback may not see committed rows.** `publish_pipeline.py:754-768` commits the version row in the request connection, then attempts S3-upload-failure rollback on a **separate connection**. Under connection pooling this rollback may not observe the just-committed row. Fix: swap the pattern for a savepoint in the same connection, or do S3 first and tolerate orphan objects (prefer the former).
2. **Anthropic judge calls have no retry.** `anthropic_client.py:64-70` uses a plain `httpx.post(..., timeout=60)` with `.raise_for_status()`. `gemini.py` already wraps calls in `gemini_request_with_retry()`; the Anthropic path should follow the same pattern so a transient 5xx doesn't silently fail the eval.
3. **`embed_texts_batch` bypasses retry.** `embeddings.py:82-121` uses a bare `httpx.Client(timeout=30)`. Should also route through `gemini_request_with_retry()` (the single-embed path already does).
4. **Tracker GitHub verification is sequential.** `tracker_service.py:108` has a `TODO` for the known-at-1k-trackers GitHub secondary-rate-limit stall. Fix with the GraphQL `repositories()` batch call (100 per request) before it becomes a production incident.
5. **`list_tracker_metrics` missing tiebreaker.** `database.py:3252` sorts by `recorded_at DESC LIMIT 50` with no `id` tiebreaker — violates the project's own "every `LIMIT` has a unique `ORDER BY` tiebreaker" rule.

### Server — code health

6. **`infra/database.py` (3,465 lines) is a god-module.** Split by domain (skills, trackers, evals, users, orgs) in a series of small mechanical moves.
7. **`domain/gauntlet.py` (1,355 lines)** should split into `patterns.py`, `credentials.py`, `shell_taint.py`, `orchestrator.py`.
8. **Five near-identical Gemini callback builders** in `publish_pipeline.py:174-277` should collapse into a single `_build_callback_fn(settings, gemini, name)` registry.
9. **`_SKILL_SUMMARY_COLUMNS` / `SkillSummary` drift risk.** `database.py:676` and `registry_routes.py:138` aren't mechanically linked — CLAUDE.md flags this as a known trap. Consider generating the Pydantic model from the column list (or vice versa) so the two can't drift.

### Client

10. **`cli/runtime.py:71,77` subprocess calls** have no `TimeoutExpired` handling and no timeout. Can hang the CLI.
11. **`registry.py:1024` skips `ziputil.validate_zip_entries`** before extraction. The helper exists in `shared/` specifically to stop zip-slip during install — but this call site doesn't call it.
12. **Token stored in plaintext** at `~/.dhub/config.{env}.json`. Worth a follow-up using `keyring` with JSON fallback.

### Frontend

13. **Duplicated skill-card markup** across `SkillsPage`, `HomePage`, and `OrgDetailPage` (~40 lines × 3). Extract a shared `<SkillCard>` component.
14. **`SkillDetailPage` zip-download lacks `AbortController`**, so rapid navigation can let a stale response overwrite state.
15. **`AskModal` uses `key={i}`** on the message list — fine today because messages are append-only, but the anti-pattern bites the next time someone adds edit/reorder.
16. **Mobile gaps.** `SkillDetailPage` sidebar is a 320px fixed column with no `@media` breakpoint; `FileBrowser` is 2-column without a stack fallback.

### Tests / observability / DX

17. **Frontend tests aren't wired into CI.** 1,661 LOC of vitest currently invisible to PR checks.
18. **No pytest markers for `integration` / `network`.** Only `slow` exists. Adding these would enable a real `make test-fast` path.
19. **No Sentry / OpenTelemetry / Prometheus**. Logging is in great shape — the gap is at the edge.
20. **Top tests to add** (ordered by leverage): auth flow end-to-end with real JWT → skill access, publish pipeline + S3 failure modes, eval timeout/cancellation, search ranking golden set, skill manifest malformed-input coverage, migration idempotence replay, frontend error-boundary behaviour, tracker state-machine races, CLI / old-server contract test.

---

## Test plan

- [x] `uv run ruff check .` → clean
- [x] `uv run ruff format --check server` → clean
- [x] `uv run --extra dev mypy` on all six edited source files → clean
- [x] `pytest server/tests/ -m "not slow"` → 943 passed, 0 failures, 34 deselected
- [x] `pytest server/tests/test_api/test_rate_limit.py server/tests/test_api/test_deps.py` → 17 passed (existing 5 + new 12)
- [x] Oversized-metadata test confirms 422 before JSON parsing

https://claude.ai/code/session_019y4YTnMTnLeugzB8fA6ENi


lfiaschi:

PR #346, opened 2026-04-22
refactor: consolidate rate-limit boilerplate + DRY/perf cleanups

lfiaschi:
## Why

A principal-engineer / architect pass over the `server/` and `client/` code surfaced a cluster of duplication and small sharp edges that were cheap to fix and unblock other work. Every change here is behaviour-preserving (or fixes a bug the old code tolerated by accident) and is covered by the existing suite plus new targeted tests — no new features, no API changes.

Net effect: **+368 / −313 LoC**, one less class of "copy-paste a 10-line closure to wire a new endpoint," and a rate limiter that stops leaking bucket entries.

## High-leverage findings from the review

The pass yielded more than what is here (see "Deferred" below), but these were picked because they're low risk, cover multiple files, and each one shrinks or speeds up a hot path.

| # | Category | Location | Change |
|---|---|---|---|
| 1 | code-health | `api/rate_limit.py` + 4 route files | One `rate_limit_dep("name")` factory replaces 9 near-identical `_enforce_*_rate_limit` closures. |
| 2 | bugs / perf | `api/rate_limit.py` | Drop `defaultdict` (empty-bucket leak) + replace O(n) purge trigger with a bounded counter. |
| 3 | code-health | `api/deps.py`, `registry_routes.py`, `tracker_routes.py` | Shared `parse_uuid` helper, stop reimplementing the same 422 path parser twice. |
| 4 | code-health | `domain/publish_pipeline.py` | Five `_build_*_fn` Gemini-wrapper factories collapse to a single `_build_gemini_fn`. |
| 5 | bugs | `api/seo_routes.py` | Cache the rendered sitemap **string**, not a shared `Response` whose headers are stamped per-call. Skill + org URLs now come from one DB query instead of two. |
| 6 | performance | `cli/registry.py` | `_ensure_tracker` reuses one `httpx.Client` (base_url + default headers) across GET/POST/PATCH — removes two extra TCP+TLS handshakes per git-based publish. |
| 7 | testing | `tests/test_api/test_rate_limit.py` | New `TestRateLimitDep` class + cases pinning down the correctness fixes in (2). |

### Rate limiter — why it needed a fix, not just a rename

The old `RateLimiter.__call__` mutated a `defaultdict(list)` and then ran `sum(len(v) for v in self._requests.values())` **on every request** as a purge trigger. Two consequences:

- Every first request from an unseen IP permanently inserted an empty list into `_requests`, even when the request was immediately accepted. Over time the map grew by one entry per unique client between purges.
- The purge trigger was O(n) per request, and `total % 100 == 0` was true every time `total == 0` (i.e. after a cold start or after a purge itself) — so the purge fired in a tight loop rather than at the intended cadence.

New code stores only buckets with at least one live timestamp, and uses a bounded counter (`_since_purge`) to schedule the sweep exactly once per `_PURGE_INTERVAL` requests. Both failure modes are now directly asserted in tests:

```python
def test_does_not_leak_empty_buckets_on_read(self) -> None: ...
def test_purge_bounds_memory_for_idle_ips(self) -> None: ...
```

### `rate_limit_dep` factory

Each route file used to carry a copy of:

```python
def _enforce_X_rate_limit(request: Request) -> None:
    state = request.app.state
    if not hasattr(state, "_X_rate_limiter"):
        state._X_rate_limiter = RateLimiter(
            max_requests=settings.X_rate_limit,
            window_seconds=settings.X_rate_window,
        )
    state._X_rate_limiter(request)
```

That's now:

```python
_rl_list_skills = rate_limit_dep("list_skills")
```

Naming convention is preserved (`settings.<name>_rate_limit` / `_rate_window`) so operators don't need to relearn any config keys, and the generated dep's `__name__` is `rate_limit_<name>` so tracebacks / profiles stay readable.

### Publish pipeline Gemini builders

Before: five ~15-line factories that each `from decision_hub.infra.gemini import <fn>`, then created a Gemini client, then returned a closure that forwarded its args to the imported function with `model=settings.gemini_model`.

After: one `_build_gemini_fn(settings, gemini, gemini_fn)` helper that does exactly that, plus five tiny one-line wrappers (`_build_analyze_fn`, `_build_analyze_prompt_fn`, …) that select the target function. The existing `test_api/conftest.py` autouse fixture that patches `_build_analyze_credential_fn` keeps working unchanged — the module-level names are preserved.

### Sitemap caching

The previous code did `cache.set("sitemap_xml", response)` where `response` is a Starlette `Response`. Starlette stamps `raw_headers` (e.g. `content-length`) at `__call__`-time, so concurrently serving the same cached `Response` to two clients is a subtle correctness trap. New code caches the XML string and builds a fresh `Response` per request, which is both cheaper to allocate and correct. The render function also builds the org URL list by deduping `row.org_slug` values from the skills query instead of firing a second `SELECT DISTINCT ...` against the same join.

### CLI `_ensure_tracker`

The previous code opened three separate `httpx.Client(timeout=60)` context managers (one per HTTP verb). Each context does a fresh TCP+TLS handshake, so a `dhub publish <git-url>` against HTTPS paid three full TLS round-trips for three API calls that all target the same host with the same headers. Now one client with `base_url=api_url, headers=headers` covers all three, and the error-handling (bail out on `typer.Exit`/426, otherwise swallow and continue) is preserved.

## Deferred (called out so they don't get lost)

Logged here for visibility — not in this PR:

- `server/src/decision_hub/infra/database.py` is 3.4k LoC mixing schema, row-mappers, CRUD, search, tracker, metrics, scan reports. Natural split: one module per domain aggregate. Risky to do in one pass; worth a follow-up.
- `RateLimiter` and `RequestLoggingMiddleware` both read `request.client.host`, which is the **proxy** IP under Modal's ingress. If/when we need honest per-client limits we'll want `X-Forwarded-For` handling with a trusted-proxy allowlist.
- `/resolve` and `/download` both bump `download_count`. Some flows (CLI: resolve → GET presigned URL) can double-count. Left alone for now since fixing it requires choosing where the "real" install event lives.
- `publish_pipeline.execute_publish` opens a fresh connection via `conn.engine.connect()` to roll back a version after a failed S3 upload. Works, but the rollback path isn't unit-tested; would be worth a fault-injection test.

## Verification

- `uvx ruff@0.15.3 check .` → clean
- `uvx ruff@0.15.3 format --check .` → 197 files formatted
- `uvx mypy client/src server/src shared/src` → 0 issues across 88 files
- `pytest` — **939 server** (incl. 11 new rate-limit cases) · **291 client** · **98 shared** — all green

## Test plan

- [x] `make lint`
- [x] `make typecheck`
- [x] `make test-server`
- [x] `make test-client`
- [x] `make test-shared`
- [ ] (reviewer) spot-check rate-limit behaviour against a deployed dev environment if desired — no setting names changed, so existing `.env.dev` values still apply.

https://claude.ai/code/session_01Ss3vrQrGhRTyuU762q7SGH


lfiaschi:

PR #348, opened 2026-04-24
fix: correctness fixes for tracker errors, LIMIT ordering, and zombie races

lfiaschi:
## Summary

Six surgical correctness fixes surfaced by a deep codebase review. No architectural refactors — every change stays inside existing module boundaries so the follow-up module splits (`database.py` 3.5k LoC, `registry_routes.py` 1.3k LoC, `client/cli/registry.py` 1.9k LoC) can land as their own focused PRs. The priority here is shipping **behavior-visible correctness wins** with focused regression tests.

### Findings addressed

| # | Fix | Why it matters |
|---|-----|-----------------|
| 1 | **Tracker partial-failure errors were silently dropped** (`tracker_service.py:804`) | When N-of-M skills failed, `last_error` was only written if *every* publish failed. Any partial failure vanished from the tracker health dashboard. `last_error` now records any failing skill; SHA-advance rule is unchanged. |
| 2 | **`has_recent_quarantine` used `LIMIT 1` without `ORDER BY`** (`database.py:2428`) | Violates the CLAUDE.md rule "every LIMIT must have an explicit ORDER BY with a unique tiebreaker." Multiple F-grade rows with matching checksum returned a non-deterministic row. Fixed: `ORDER BY created_at DESC, id DESC`. |
| 3 | **`fetch_similar_skills` top-N had no tiebreaker** (`database.py:2128`) | Near-duplicate embeddings with identical `vec_dist` produced a non-deterministic top-5, so the same user could see different "similar skills" on refresh. Fixed: tiebreaker on `skills.id ASC`. |
| 4 | **`_refresh_skill_latest_version` ordered only by semver triple** (`database.py:1384`) | The `(skill_id, semver)` unique constraint *should* make this unnecessary, but belt-and-braces: added `versions.id DESC` tiebreaker. |
| 5 | **Zombie-heartbeat race could clobber a live worker** (`registry_routes.py:1100`) | `_check_zombie` did a read-check-update sequence: read heartbeat, compare to wallclock, unconditionally `UPDATE status='failed'`. A worker sending a fresh heartbeat in that window would be overwritten. Introduced `mark_eval_run_failed_if_stale()` — an atomic conditional `UPDATE` gated on the observed `heartbeat_at` value. If a live worker updates `heartbeat_at` first, our `UPDATE` affects 0 rows and the run is left running. |
| 6 | **Duplicate SKILL.md parsing in tracker removal detection** (`tracker_service.py:857`) | `_detect_removed_skills` re-parsed every manifest that `discover_skills` had already parsed. Added `discover_skills_with_names()` → `list[tuple[str, Path]]`; the tracker threads the pre-parsed names through, halving the parse work per tracker tick. |

### Test coverage

New tests (all SQL-shape or behavior assertions, no DB required):
- `test_has_recent_quarantine` — deterministic `ORDER BY … LIMIT 1`
- `test_fetch_similar_skills` / `test__refresh_skill_latest_version` — SQL-shape tiebreaker assertions (new `test_limit_ordering.py`)
- `test_mark_eval_run_failed_if_stale` — returns `False` when `rowcount=0`, `UPDATE` gated on observed heartbeat + active-worker status set (new `test_mark_eval_run_failed_if_stale.py`)
- `test_zombie_detection_respects_concurrent_heartbeat` — end-to-end: if atomic update returns False, the endpoint does *not* report `failed`
- `test_partial_success_advances_sha_and_records_errors` — partial-failure errors appear in `last_error` (previously asserted `is None`)
- `test_three_of_five_succeed_advances_sha_and_records_errors` — multi-skill variant of the same guarantee
- `test_discover_with_names_returns_name_and_path` + `test_discover_parses_each_manifest_exactly_once`

Updated existing tests for the new tuple-return and patched-function names.

### Test plan

- [x] `uvx ruff check .` → clean
- [x] `uvx ruff format --check .` → clean
- [x] `uvx mypy client/src/ server/src/ shared/src/` → 0 issues
- [x] server: `pytest server/tests/ -m "not slow"` → 942 passed, 34 deselected
- [x] client: `pytest client/tests/` → 291 passed, 29 skipped
- [x] shared: `pytest shared/tests/` → 98 passed

### Deliberately out of scope (follow-up PRs)

Flagged during the review but not addressed here to keep the PR tight:
- Module splits: `database.py` (3.5k LoC) → per-domain submodules; `registry_routes.py` (1.3k LoC) → sub-routers; `client/cli/registry.py` (1.9k LoC) → command modules.
- Backfill scripts not resumable/checkpointed (CLAUDE.md §97 violation).
- Duplicate secret-redaction regex across `evals.py` and `anthropic_client.py`.
- `_log_ask_analytics` swallow-and-log has no metric/alert hook.

https://claude.ai/code/session_01JVy1BKgfcZSBH9cPBkNFYz

---
_Generated by [Claude Code](https://claude.ai/code/session_01JVy1BKgfcZSBH9cPBkNFYz)_

### DIFF
--- a/server/src/decision_hub/api/registry_routes.py
+++ b/server/src/decision_hub/api/registry_routes.py
@@ -58,9 +58,9 @@ from decision_hub.infra.database import (
     list_granted_skill_ids,
     list_skill_access_grants_with_names,
     list_user_org_ids,
+    mark_eval_run_failed_if_stale,
     resolve_latest_version,
     resolve_version,
-    update_eval_run_status,
     update_skill_visibility,
 )
 from decision_hub.infra.database import (
@@ -1100,8 +1100,13 @@ def _run_to_response(run) -> EvalRunResponse:
 def _check_zombie(conn: Connection, run) -> str:
     """Check if a running eval run has a stale heartbeat (zombie).
 
-    If heartbeat_at is older than _STALE_HEARTBEAT_SECONDS, marks the
-    run as failed and returns "failed". Otherwise returns run.status.
+    If heartbeat_at is older than _STALE_HEARTBEAT_SECONDS, atomically
+    marks the run as failed (gated on the heartbeat not moving in the
+    meantime) and returns "failed". Otherwise returns run.status.
+
+    The UPDATE is conditional on ``heartbeat_at`` still equalling the
+    value we observed, so a live worker emitting a fresh heartbeat
+    between the read and the write is not clobbered.
     """
     if run.status not in ("running", "judging", "provisioning"):
         return run.status
@@ -1109,14 +1114,15 @@ def _check_zombie(conn: Connection, run) -> str:
         return run.status
     elapsed = (datetime.now(UTC) - run.heartbeat_at).total_seconds()
     if elapsed > _STALE_HEARTBEAT_SECONDS:
-        update_eval_run_status(
+        marked = mark_eval_run_failed_if_stale(
             conn,
             run.id,
-            status="failed",
+            observed_heartbeat_at=run.heartbeat_at,
             error_message=f"Stale heartbeat ({int(elapsed)}s). Worker may have crashed.",
             completed_at=datetime.now(UTC),
         )
-        return "failed"
+        if marked:
+            return "failed"
     return run.status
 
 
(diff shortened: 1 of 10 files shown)


lfiaschi:

PR #363, opened 2026-05-05
refactor: high-leverage cleanups from a deep codebase review

lfiaschi:
## Why this PR

This is the second half of a two-step "review then refactor" pass over
the monorepo. The review surfaced ~80 candidate findings across
architecture, code-health, bugs, security, performance, and testing.
After verification (several agent-surfaced findings turned out to be
false positives — see "What we deliberately *didn't* change" below),
five fixes hit the sweet spot of **clearly high-leverage, low-risk,
testable in isolation, and small enough to review in one PR**.

The intent is to land these now, then follow up with the larger
structural refactors (centralised skill-access query layer, splitting
`execute_publish` into prepare/commit/trigger, tracker state machine)
in dedicated PRs.

### 1. Rate-limiter factory dedup — `api/rate_limit.py`, `api/registry_routes.py`, `api/search_routes.py`
**Before:** 8 nearly-identical `_enforce_*_rate_limit` functions, each
~10 lines, duplicating the same lazy-init-on-`app.state` pattern.
Adding a new endpoint group meant copy-pasting one of them and
remembering to change three names.

**After:** one `make_rate_limit_dep(name)` helper in `rate_limit.py`
that reads `{name}_rate_limit` / `{name}_rate_window` off `Settings`
and caches the limiter under `app.state._rate_limiter__{name}`. Each
endpoint group is now one line:

```python
```

Net: −80 LOC, identical runtime behaviour, easier to add new groups.

### 2. Eval-report endpoint body dedup — `api/registry_routes.py`
The query-string variant (`/skills/{org}/{name}/eval-report?semver=`)
and the path variant (`/skills/{org}/{name}/versions/{semver}/eval-report`)
had byte-identical bodies. Both URLs are in active use (frontend uses
the query-string form, CLI uses the path form), so the routes stay —
but both delegate to one `_resolve_eval_report` helper now. Future
visibility-check changes touch one place instead of two.

### 3. CLI config file is now `0o600` — `client/src/dhub/cli/config.py`
**The bug:** `~/.dhub/config.{env}.json` holds a year-long JWT but
`save_config` wrote it via `Path.write_text`, which uses the process
umask (typically 022 on Linux → world-readable). On a shared host or
CI machine, any local user could `cat` the file and impersonate the
owner.

**The fix:** open the file via `os.open(path, O_WRONLY|O_CREAT|O_TRUNC,
0o600)` so it's never briefly world-readable mid-write (TOCTOU on
shared hosts). After writing, re-`chmod` the file *and* the parent
`~/.dhub` to `0o700` in case either pre-existed with looser perms — a
user who logged in before this change would otherwise stay exposed
forever. `os.chmod` failures (FAT/network mounts) are suppressed:
best-effort tightening is still better than nothing.

### 4. Reject symlink entries in zip extraction — `shared/src/dhub_core/ziputil.py`
**The bug:** the existing path-traversal check only validated entry
*names*. A zip entry archived as a Unix symlink (`data.json` →
`/etc/passwd`) has a safe-looking name, passes the existing check,
gets extracted, and then host files leak the moment any later code
does `open(skill_dir / "data.json")`.

**The fix:** `_is_symlink(info)` checks the top 16 bits of
`external_attr` for `S_IFLNK` (0o120000) and `validate_zip_entries`
rejects all symlinks outright. Skill packages have no legitimate
use for symlinks. Windows-created archives leave those mode bits
zero and remain unaffected — there's an explicit regression test
for that.

### 5. `/ask` thread pool is now a module-level singleton — `api/search_routes.py`
**Before:** every `/ask` request built a fresh
`ThreadPoolExecutor(max_workers=2)` to overlap guard+embed, paying
~thread-spawn cost per call.

**After:** one `_ASK_PARALLEL_POOL` at module scope, reused across
requests. Workers are named (`ask-parallel`) for stack-trace
debuggability. Same parallelism, no per-request overhead.

## Tests added (10 new tests, all behavioural)

- `shared/tests/test_ziputil.py` — symlink rejection (basic + with
  traversing name + Windows-zero-mode regression guard)
- `client/tests/test_cli/test_config.py` — `save_config` produces a
  `0o600` file; `save_config` tightens an existing loose-mode file
- `server/tests/test_api/test_rate_limit.py` — `TestMakeRateLimitDep`
  class: limiter is cached on `app.state` (not rebuilt per request),
  distinct names get distinct limiters, 429 still trips at the
  configured ceiling
- `server/tests/test_api/test_search_routes.py` — `/ask` reuses the
  same `_ASK_PARALLEL_POOL` instance across two consecutive requests
  and never gets shut down between them

## Quality gates

| Gate | Result |
|---|---|
| `ruff check` | ✅ All checks passed |
| `ruff format --check` | ✅ 197 files already formatted |
| `mypy client/src server/src shared/src` | ✅ No issues found in 88 source files |
| `pytest shared/tests` | ✅ **101 passed** |
| `pytest client/tests` | ✅ **293 passed**, 29 skipped (docx integration, env-gated) |
| `pytest server/tests -m "not slow"` | ✅ **937 passed**, 34 deselected (slow LLM regression) |

## What we deliberately *didn't* change (and why)

The review surfaced these candidates that turned out to be wrong on
inspection — leaving them here so reviewers can see they were
considered:

- **Sitemap "O(n²) string concat"** — `seo_routes.py` already uses
  `lines.append(...)` + `"\n".join(lines)`, which is O(n).
- **Search analytics "engine leak"** — `_log_ask_analytics` is passed
  the *shared* app engine and uses `engine.begin()`; calling
  `engine.dispose()` would tear down the application's connection
  pool. The current code is correct.
- **Zombie eval `get_eval_run` "doesn't use updated record"** — line
  1136 already does `run = find_eval_run(conn, parsed_id)` after
  `_check_zombie`, so the response sees the post-zombie status.
- **Delete a duplicate eval-report endpoint** — both URLs have active
  callers (frontend vs CLI), so we deduped the *body* via a helper
  instead.

## Follow-up work (not in this PR)

In rough priority order, surfaced by the review and worth their own PRs:

- Centralise skill access-control in the query layer
  (`fetch_skill_for_user(...)`) — currently scattered across ~14
  endpoints, each one a potential auth bypass on the next refactor.
- Split `domain/publish_pipeline.py:execute_publish` (264 LOC, 15 args)
  into `prepare → commit → trigger` so domain logic is testable
  without S3/Gemini mocks and so S3-upload failures can't leave
  orphaned DB rows.
- Extract a `TrackerStateMachine` from `domain/tracker_service.py` so
  cron can detect and release stuck "claimed but not progressing"
  trackers.
- Address `_refresh_skill_latest_version` race window (concurrent
  publishes of the same skill can briefly write a non-monotonic
  `latest_semver`).
- Add a background sweeper for zombie eval runs — currently zombies
  are only marked failed when a user happens to poll the run.

## Test plan

- [x] Unit tests pass on all four packages
- [x] Lint + format clean
- [x] Mypy clean
- [ ] Smoke-test `dhub login` on dev to confirm `~/.dhub/config.dev.json`
  lands as `0o600` (deferred to whoever picks this up — the unit test
  asserts the same property)
- [ ] Manual: hit `/v1/ask` twice and confirm only two threads named
  `ask-parallel-0` / `ask-parallel-1` ever exist (deferred)

https://claude.ai/code/session_01Xt211PAu27pzbLxipJX9af

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xt211PAu27pzbLxipJX9af)_


lfiaschi:

PR #367, opened 2026-05-10
refactor(server): dedupe rate-limit boilerplate and Gemini JSON parsing

lfiaschi:
## Summary

Consolidates two long-standing API-layer DRY violations behind small, well-tested helpers. Both refactors are pure code-organisation wins, but each also closes a small correctness gap that had been hiding in the duplicated copies.

This PR is the implementation half of a deeper review pass over the codebase; the highest-leverage findings that did **not** ship in this PR are summarized at the bottom for follow-up issues.

### 1. Rate-limiter dependencies

**Before** (`registry_routes.py` x7, `search_routes.py` x1, `auth_routes.py` x1 — 9 copies):

```python
def _enforce_search_rate_limit(request: Request) -> None:
    state = request.app.state
    if not hasattr(state, "_search_rate_limiter"):
        state._search_rate_limiter = RateLimiter(
            max_requests=settings.search_rate_limit,
            window_seconds=settings.search_rate_window,
        )
    state._search_rate_limiter(request)
```

The lazy-init pattern carried a subtle race: two concurrent first-requests could each pass the `hasattr` check, each build a fresh `RateLimiter`, and the second one's assignment would silently throw away the first one's counter.

**After** — one declarative registry, one factory, one tiny dependency:

```python
# rate_limit.py
_RATE_LIMIT_NAMES = ("search", "list_skills", "resolve", ..., "scan_report")

def build_rate_limiters(settings) -> dict[str, RateLimiter]: ...
def rate_limit(name: str): ...   # FastAPI dependency

# app.py
app.state.rate_limiters = build_rate_limiters(settings)

# routes
_rl_publish = rate_limit("publish")

@router.post("/publish", dependencies=[Depends(_rl_publish)])
def publish_skill(...): ...
```

- Eager construction in `create_app()` removes the lazy-init race entirely.
- A typo in a route's name (`rate_limit("publsh")`) raises `KeyError` at module import time — never silently in prod.
- Test apps that bypass `create_app` still work: the dependency lazy-builds the dict from `app.state.settings` on first call.

### 2. Gemini JSON-array judgment parsing

`analyze_code_safety`, `analyze_credential_entropy`, and `analyze_prompt_safety` had ~30 lines each of nearly-identical `try / json.loads / for-each model_validate / except: pass` plumbing. The `except json.JSONDecodeError: pass` arm swallowed the underlying cause silently — production was blind to LLM regressions and to our own prompt-template breakage.

Extracted to `_parse_judgment_array(text, schema, item_fallback, log_label, skill_name)`:

- Returns a list of validated judgments (with caller-supplied fail-closed dicts substituted for any item that fails schema validation), **or**
- Returns `None` to signal a whole-array fallback (non-JSON, non-list root, etc.) — caller decides the fallback shape since it varies per check.
- JSON decode errors and non-list roots are now logged at warning level with the skill name and the underlying cause.

The credential-entropy post-processing (merging `label`/`line` from input hits onto the LLM's index-keyed response) is preserved unchanged.

## Test plan

- [x] `pytest server/tests -m 'not slow'` — **945 passed, 34 deselected**
- [x] `pytest client/tests` — **291 passed, 29 skipped** (skips pre-existing)
- [x] `pytest shared/tests` — **98 passed**
- [x] `ruff check . && ruff format --check .` — clean
- [x] `mypy client/src server/src shared/src` — clean (88 source files)
- [x] 7 new unit tests in `test_rate_limit.py` covering the factory and dependency
- [x] 5 new unit tests in `test_gemini.py` covering the parse-helper, including a regression guard that JSON-decode errors are logged (the previous `except: pass` would silently fail this test)

No behaviour change at the route level: same 429 responses, same fail-closed `dangerous=True` semantics, same logger message stems.

## Out of scope (follow-up candidates from the review)

These were identified during the review but deliberately deferred to keep this PR focused and reviewable:

1. **Split `infra/database.py` (3,465 LOC)** by table family (`db/skills.py`, `db/orgs.py`, `db/audit.py`, `db/evals.py`).
2. **Add direct unit tests** for `seo_routes`, `registry_service`, `domain/orgs.py`, `domain/repo_utils.py`, `domain/crypto.py` (currently only indirectly covered).
3. **Reduce `jwt_expiry_hours` default** from 8760 (1 year) to 24h with a refresh flow.
4. **Replace manual retry loops** in `gemini.py` with a single decorator (e.g. `tenacity`).
5. **Tighten the `except Exception` arms** around the Cisco scanner bridge in `publish_pipeline.py`.

Each of these is a self-contained change — happy to file individual issues.

https://claude.ai/code/session_011XjsPw1vGwq56q6bEdgeXg

---
_Generated by [Claude Code](https://claude.ai/code/session_011XjsPw1vGwq56q6bEdgeXg)_


lfiaschi:

PR #371, opened 2026-05-15
refactor: tighten CLI subprocess timeouts, server query determinism, frontend type scale

lfiaschi:
## Context

This PR is the output of a deep architect/principal-engineer pass over
the codebase. The review surfaced ~12 high-leverage findings; this PR
ships the **6 small, verified, low-risk fixes** that are aligned to
explicit project rules in `CLAUDE.md`. The remaining findings are
flagged at the bottom for follow-up PRs.

## Summary

| # | Area | Change |
|---|------|--------|
| 1 | CLI / `client/src/dhub/core/git_repo.py` | Bound every `git` subprocess with a 120s timeout; clean up tmpdir on any failure |
| 2 | CLI tests / `client/tests/test_core/test_git_repo.py` | First test coverage for `clone_repo` — timeout, cleanup, SHA-ref path |
| 3 | Server / `infra/database.py:2128` | Add `skills.id` as unique tiebreaker on the vector-similarity ORDER BY |
| 4 | Server / `scripts/backfill_embeddings.py:57` | Add ORDER BY to the embedding-NULL batch fetch |
| 5 | Frontend / `pages/NotFoundPage.tsx` | Replace hardcoded `fontSize` with `var(--text-hero)` / `var(--text-heading)` |
| 6 | Frontend / `pages/OrgDetailPage.tsx` | Same — both 404 and generic error branches |

## Why each change

**1. CLI: bound git subprocess time** — `subprocess.run` for `git clone`
and `git checkout` had no `timeout=`. A slow remote, hung credential
helper, or DNS hang would freeze the CLI indefinitely with no Ctrl-C-able
intermediate state. Extracted a `_run_git` helper that wraps every git
invocation at one place (120s bound, `TimeoutExpired` → `RuntimeError`)
and moved tmpdir cleanup into a `try/except` around the whole flow so
a failure in the second invocation (the SHA-checkout path) can no longer
leak `dhub-repo-*` directories under tempdir.

**2. CLI tests** — `clone_repo` had zero coverage; only `discover_skills`
was tested. Added four tests that exercise the new behavior: timeout
kwarg is always passed, `TimeoutExpired` surfaces as `RuntimeError`,
failed clones don't leak tmp dirs, SHA refs do clone+checkout.

**3. Server: vector similarity tiebreaker** —
`find_similar_skills_by_vector` ordered only by `vec_dist ASC`.
`CLAUDE.md` explicitly states *"every query with `LIMIT` must have an
explicit `ORDER BY` with a unique tiebreaker"*. Ties at the same cosine
distance could swap positions across calls, making the "similar skills"
panel flicker. Adding `skills_table.c.id` as a stable tiebreaker is a
one-line, behavior-preserving fix.

**4. Server: backfill embeddings ORDER BY** — Same project rule. The
fetch-next-batch query at line 57 had `.limit(batch_size)` with no
`ORDER BY`. The set shrinks each batch so the practical bug is small,
but the rule is explicit and adding `order_by(skills_table.c.id)` is
free.

**5–6. Frontend type scale** — `NotFoundPage.tsx` and `OrgDetailPage.tsx`
had inline `fontSize: "4rem"` / `"1.6rem"` for the 404 and error states.
`CLAUDE.md` mandates the `--text-*` scale variables so that the mobile
breakpoint at `480px` can step every size down one level. Swapped to
`var(--text-hero)` (4rem) and `var(--text-heading)` (1.6rem).

## Verification

- `make test-client` — **295 passed**, 29 skipped (4 new tests for `clone_repo`)
- `make test-shared` — **98 passed**
- `cd frontend && npx vitest run` — **82 passed**
- `make lint` — clean (ruff check + format)
- `cd frontend && npx tsc --noEmit` — clean
- Modified server modules import cleanly under `uv run --package decision-hub-server`

## Out of scope (deferred to follow-up PRs)

These findings came up in the review but were too risky or too broad to
include here without a more targeted plan:

- **`api/registry_service.py:214-263`** — assessment failure path
  constructs **two fresh `Engine` instances** inside the exception
  handler. During a DB outage this multiplies pool churn at exactly the
  moment the DB can least afford it. Should reuse the request-scoped
  engine.
- **`domain/tracker_service.py:46-74`** — sequential `gh.get()` per URL
  in a Python loop; the existing `TODO` at line 108 calls out the
  GitHub secondary-rate-limit risk at 1000+ URLs.
- **`cli/registry.py:427, 449, 471, 486`** — `except Exception: pass`
  for git branch detection and tracker calls. At minimum these should
  `logger.debug(..., exc_info=True)` so debug-mode users can see why a
  best-effort path failed.
- **`scripts/crawler/__main__.py` and `scripts/tracker_health.py`** —
  still use `print()` instead of `loguru.logger`; production logs are
  unstructured stdout from these scripts.
- **CLI HTTP boilerplate** — `with httpx.Client(timeout=60) as client:`
  is repeated ~34 times across the CLI; a centralized factory in
  `cli/config.py` would DRY this up and make timeout tuning a one-line
  change.

## Note on `uv.lock`

The diff includes `uv.lock` because `dhub-cli` was at `0.12.3` in
`client/pyproject.toml` but `0.12.1` in the lockfile (and `dhub-core`
`0.2.10` vs `0.2.8`). Running `uv run` to verify the server imports
reconciled the lockfile to match the pyproject. Reverting it would just
make the lockfile inconsistent again on the next `uv` invocation.

https://claude.ai/code/session_01VQvH3NJVk3JqJPVw1SYK7M

---
_Generated by [Claude Code](https://claude.ai/code/session_01VQvH3NJVk3JqJPVw1SYK7M)_


lfiaschi:

PR #373, opened 2026-05-18
fix: review-driven hardening (CLI config 0600, GH-App token cache, useCountUp, error-body cap)

lfiaschi:
## Summary

A codebase-wide architecture + principal-engineer review (full notes at the bottom) surfaced ~10 high-leverage issues. This PR ships the smallest set of focused fixes that are independently verifiable, regression-tested, and risk-bounded — one real security bug, one efficiency win that reuses an existing primitive, one frontend bug, and one defensive cleanup, each with tests. The remaining findings are itemized below as follow-ups.

### 1. CLI config saved 0o600 — `client/src/dhub/cli/config.py:82-117`
`save_config()` previously wrote `~/.dhub/config.{env}.json` with default umask permissions (typically `0o644`), so the auth token inside was readable by every other user on a shared host. The file is now created in one syscall with `os.open(..., O_WRONLY|O_CREAT|O_TRUNC, 0o600)`. Pre-existing loose files (legacy installs) are tightened on the next save. `CONFIG_DIR` is also chmodded to `0o700`. All `chmod` calls are wrapped in `contextlib.suppress(OSError, NotImplementedError)` for non-POSIX filesystems.

- New test `test_saved_config_is_owner_only` asserts owner-only mode bits.
- New test `test_save_tightens_pre_existing_loose_permissions` asserts legacy 0o644 files are tightened.

### 2. Cache GitHub App installation tokens — `server/src/decision_hub/infra/github_app_token.py`
`mint_installation_token` was called on every tracker tick (`tracker_service.py:106, 252, 710`) and every Modal entrypoint (`modal_app.py:157, 198`) — each call is a real HTTPS round-trip to GitHub plus a JWT sign. GitHub installation tokens live ~60 minutes, so caching is the obvious win.

Implementation reuses the existing `TTLCache` from `infra/cache.py` (no new dependency, already thread-safe and well-tested) keyed by `(app_id, installation_id)` with a 50-minute TTL — short enough to never serve a token close to expiry, long enough to collapse the per-tick mint into a single call per container per ~50 min. Added a `use_cache=False` escape hatch and a `clear_token_cache()` helper for tests.

- New tests for cache hits, per-installation isolation, `use_cache=False` bypass, and `clear_token_cache()` forced re-mint.

### 3. `useCountUp` re-animates when target changes — `frontend/src/hooks/useCountUp.ts`
The `hasAnimated` ref was set to `true` on first animation and never cleared. Once the stats API returned a value, the counter froze on it forever — if the stats refetched to a new number, the UI showed the old one. The animation also had no `cancelAnimationFrame` cleanup on unmount.

Fix tracks the last animated target in a separate `animatedFor` ref; when `target` changes to a new non-zero value, the latch is released so the new value animates. The `rafId` is now cancelled on unmount.

- New `useCountUp.test.ts` covers the gating, the re-animation, and absence of throws across target transitions.

### 4. `fetchJSON` caps error body — `frontend/src/api/client.ts`
Previously non-OK responses inlined the entire response body into `Error.message`. When the upstream proxy returned an HTML 500 page, the message could be many KB of HTML — noisy in dev tools, ugly in the UI, and unbounded in memory. Now capped at 1 KiB with a `… (N more chars truncated)` suffix; empty bodies report `(empty response)` instead of `API 500: `.

- Two new error-handling tests verify truncation and empty-body messaging.

### 5. `uv.lock` refresh
Drift fix: `client/pyproject.toml` declared `0.12.3` and `shared/pyproject.toml` declared `0.2.10`, but `uv.lock` was still pinned to the prior versions. `uv run` synced it; included in this PR so CI matches local.

## Test plan

- [x] `uvx ruff format --check .` — clean (197 files)
- [x] `npx tsc -b` (frontend) — clean
- [x] `npx eslint` on changed frontend files — clean
- [x] `pytest client/tests/` — 293 passed, 29 skipped
- [x] `pytest server/tests/test_infra/ -m "not slow"` — 168 passed
- [x] `pytest server/tests/test_infra/test_github_app_token.py` — 8 passed (was 4)
- [x] `pytest client/tests/test_cli/test_config.py` — 13 passed (was 11)
- [x] `npx vitest run` — 87 passed across 9 files (added `useCountUp.test.ts` + new error-handling cases)

## Review findings deferred to follow-ups

Triaged but **not** included to keep this PR reviewable. Each is independently scoped:

| # | Title | Cat. | Impact | Effort | Citation |
|---|---|---|---|---|---|
| 6 | `tracker_service._verify_repos_removed` not batched at 1000+ URLs (GitHub secondary rate-limit) | Reliability | M | M | `tracker_service.py:108` |
| 7 | `storage.py` S3 ops rely on boto3 default retries; make explicit `Config(retries=...)` for visibility | Reliability | L | S | `infra/storage.py:43` |
| 8 | `SkillDetailPage` retry mechanism re-triggers `useEffect` indirectly via `setZipLoading(false)` — works, but fragile | Code health | L | M | `frontend/src/pages/SkillDetailPage.tsx:200-216` |
| 9 | Hardcoded inline `fontSize` strings in 404 fallbacks bypass the type-scale CSS vars | DX | L | S | `OrgDetailPage.tsx:62-63,70`, `SkillDetailPage.tsx:254` |
| 10 | No tests for `seo_routes`, `activate_trackers.py`, or `validate_jwt()` non-mocked paths | Testing | M | M | (per the test/CI audit) |
| 11 | CI lacks Python dependency caching; `uv run` re-downloads each job | CI speed | L | S | `.github/workflows/ci.yml` |

## Review highlights (Step 1)

**System map.** uv-workspace monorepo: `client/` (open-source Typer CLI → PyPI), `server/` (FastAPI on Modal), `shared/` (`dhub-core` — single source of truth for `SkillManifest`/`RuntimeConfig`/manifest parsing), `frontend/` (React 19 + Vite, bundled into the server image at deploy). ~54k LoC Python + ~6.7k TS/TSX. Layering is clean (`api → domain → infra`).

**Architectural health: strong.** Per-IP sliding-window rate limits, 30s DB statement timeout, JWT carries `github_orgs` so per-request DB lookups are avoided, DB-commit-before-S3 on publish with rollback (`publish_pipeline.py:750-768`), fail-open on non-critical paths (embeddings, scan reports, eval triggers), secret redaction in `logging.py:64-80`, security-headers middleware via raw ASGI (avoids `BaseHTTPMiddleware` receive-wrapping deadlock), zip-slip validation actually used at install time (`registry.py:1027-1033`).

**Top 10 high-leverage changes.**

| # | Title | Cat. | Impact | Effort | Status |
|---|---|---|---|---|---|
| 1 | CLI config world-readable (token leak) | Sec | H | S | **fixed in this PR** |
| 2 | GH-App token re-minted on every poll | Perf/quota | M | S | **fixed in this PR** |
| 3 | `useCountUp` cannot replay | Bug | M | S | **fixed in this PR** |
| 4 | `fetchJSON` bleeds upstream HTML into UI | UX/health | M | S | **fixed in this PR** |
| 5 | `test_config.py` doesn't assert file mode | Testing | M | S | **fixed in this PR** |
| 6 | No tests for token caching | Testing | M | S | **fixed in this PR** |
| 7 | `_verify_repos_removed` not batched | Reliability | M | M | follow-up |
| 8 | `SkillDetailPage` retry indirection | Health | L | M | follow-up |
| 9 | S3 retries implicit only | Reliability | L | S | follow-up |
| 10 | Inline `fontSize` strings bypass the scale | DX | L | S | follow-up |

Open question: the `publish_pipeline.py:756-768` S3 rollback uses a fresh DB connection — if the request drops after upload starts but before rollback finishes, the audit log will reflect the published version even though the S3 object never landed. Worth measuring in prod traces; not changed here because the right fix may be a request-level timeout, which crosses several layers.

https://claude.ai/code/session_01WqqAp4xyiH2boGZuRVCagQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01WqqAp4xyiH2boGZuRVCagQ)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its GitHub App installation-token caching is catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #374, opened 2026-05-19
refactor: harden rate limiter, S3 paging, manifest path, config perms

lfiaschi:
## Summary

A multi-agent codebase review surfaced eight concrete issues that were each (a) small, (b) high-impact, and (c) testable in isolation. Bundling them into one PR so the review burden is the surface area, not the count.

Every change is covered by tests. All quality gates pass locally: ruff + format + mypy clean; 970 server tests, 296 client tests, 103 shared tests, 84 frontend tests; `tsc` + ESLint clean.

### Server

- **Rate limiter — TOCTOU race + behind-proxy keying**
  `api/rate_limit.py`, `api/registry_routes.py`, `api/search_routes.py`, `api/auth_routes.py`
  - Eight near-identical `_enforce_*_rate_limit` helpers used `hasattr` + `setattr` to lazily create limiters on `app.state`. Two concurrent first-requests on a fresh container could both pass the `hasattr` check and stomp each other, splitting the counter across independent buckets and leaking requests over the cap. Limiters are now built up front in `create_app()` via `register_rate_limiters(app, settings)`; routes use `Depends(rate_limited("name"))` — one factory, no race.
  - The limiter keyed on `request.client.host`, which behind Modal's HTTPS edge is always the proxy IP — every caller shared one bucket. New `trust_forwarded_for` setting (default `true`) honours the leftmost `X-Forwarded-For` entry as the real client.
  - Per-IP timestamp store moved from list-rebuild (O(n) per call) to `deque` with `popleft` (amortised O(1)). Stale-key pruning is now triggered by a per-call counter instead of `total % 100 == 0`, which fired unpredictably when the dict shrank.

- **`get_current_user` / `_optional` duplicated decode**
  `api/deps.py`
  Two near-identical functions doing JWT decode + legacy-claim handling. Factored a private `_decode_request_token` so they share one definition of "valid"; new unit tests guard against the two paths drifting again.

- **DoS: unbounded publish metadata + traversal in `manifest_path`**
  `api/registry_routes.py`
  - `metadata: str = Form(...)` had no `max_length` — a 100 MB JSON body could be uploaded. Capped at 10 KiB.
  - `manifest_path` was stored verbatim from publisher input. New `_validate_manifest_path` rejects `..`, leading slash/backslash, control chars, and length > 512. This field is surfaced in API responses and "view on GitHub" links.

- **S3 pagination bug — eval logs silently truncated**
  `infra/storage.py`
  `list_eval_log_chunks` and `delete_eval_logs` issued a single `list_objects_v2` call. Any eval that produced more than ~1000 chunks lost everything past the first page; deletes leaked objects. Both now use the boto3 paginator; deletes batch in groups of 1000 (the S3 cap).

### Shared (`dhub-core`)

- **Manifest body DoS**
  `dhub_core/manifest.py`
  `parse_skill_md` called `path.read_text()` with no size check. A multi-megabyte body would OOM PyYAML, blow up LLM token budgets, and get persisted to Postgres. New 2 MiB ceiling enforced via `stat()` before the read.

- **Zip symlink entries**
  `dhub_core/ziputil.py`
  `validate_zip_entries` checked path traversal but not symlinks. Python's stdlib materialises symlink entries as plain files, but downstream tooling (git, system unzip) creates real links and can then be steered outside the sandbox. UNIX-mode symlink entries are now rejected outright.

### Client (`dhub-cli`)

- **`~/.dhub/config.*.json` written world-readable**
  `cli/config.py`
  The file holds a long-lived JWT bearer token; under the default umask it was readable by any local user on shared hosts. `save_config` now `chmod`s the file to `0o600` after write (best-effort: filesystems that don't support chmod don't fail the login).

- **`dhub init` foot-guns**
  `cli/init.py`
  - Bare `typer.prompt` accepted Enter for description, then manifest validation failed later at publish time with a confusing round-trip. New `_prompt_nonempty` loops until the user supplies a real value.
  - Descriptions were string-interpolated into `description: "{value}"`, producing invalid YAML for any input containing a `"`. Now goes through `yaml.safe_dump` so quotes/colons round-trip safely.
  - `validate_skill_name` failures showed a Python traceback. Now wrapped in `try/except ValueError` and rendered as a friendly error.

### Frontend

- **Raw server errors leaked to the UI**
  `api/client.ts`
  `fetchJSON` threw `new Error('API 500: ' + raw body)`, which `useApi` then rendered straight into the UI — internal hostnames and stack frames included. New `ApiError` carries the status alongside a sanitised user-facing message (mapped per status class); raw 5xx bodies are still logged to the console for dev tools.

- **Stale closure on rapid refetch**
  `hooks/useApi.ts`
  `refetch()` and the auto-fetch effect used independent `cancelled` flags, so a manual refetch issued mid-flight could be overwritten by a stale auto-fetch response. Single monotonic fetch-id now governs both paths.

- **`GradeBadge` broke the typographic scale**
  `components/GradeBadge.module.css`
  Hardcoded `font-size: 0.65rem / 0.8rem / 1.4rem`. Replaced with `var(--text-xxs / --text-xs / --text-lg)` so the badge tracks the scale defined in `index.css`.

- **Duplicate `REMARK_PLUGINS` declaration**
  `constants/markdown.ts` (new), `components/AskModal.tsx`, `pages/SkillDetailPage.tsx`
  Two files independently declared `const REMARK_PLUGINS = [remarkGfm]`. Centralised so the plugin set stays in lockstep across surfaces that render user-controlled markdown.

## What this PR does NOT do

Findings from the review that need their own focused PRs and are deliberately out of scope here:

- `database.py` (3,465 LOC) split. Critical to do, large mechanical churn — better as its own diff.
- LLM prompt-injection hardening in `serialize_index`. Needs a design discussion about whether to wrap data in delimiters or strip suspicious leading text from descriptions.
- ThreadPoolExecutor created per-request in `ask_skills`. Wants a global bounded pool — touches request lifecycle.
- Skill descriptions / publish-pipeline coverage gaps in tests. Listed for follow-up.

## Test plan

- [x] `make lint` (ruff check + format check)
- [x] `uvx mypy client/src/ server/src/ shared/src/` — clean
- [x] `pytest server/tests/ -m "not slow" --no-cov` — 970 passed
- [x] `pytest client/tests/` — 296 passed, 29 skipped (docx)
- [x] `pytest shared/tests/` — 103 passed
- [x] `npx vitest run` (frontend) — 84 passed
- [x] `npx tsc -b && npx eslint .` — clean (one pre-existing exhaustive-deps warning, untouched)
- [x] Added 59 new tests targeting every behaviour change above; old tests for affected modules either still pass unchanged or were updated to match the new contracts (e.g. paginated S3 list stubs, registered limiters on test apps).

https://claude.ai/code/session_01MpWQTvbqvQHXTwjiLsewki

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpWQTvbqvQHXTwjiLsewki)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -82,7 +83,12 @@ def load_config() -> CliConfig:
 def save_config(config: CliConfig) -> None:
     """Save CLI config to ~/.dhub/config.{env}.json.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ~/.dhub directory if it does not already exist and locks
+    the resulting file to owner-only read/write (``0o600``). The file
+    contains a long-lived JWT bearer token; without explicit perms it
+    inherits umask 0o022, making the token world-readable on any shared
+    host. ``chmod`` is a no-op on Windows, where ACLs already restrict
+    files in the user's home to that user.
     """
     CONFIG_DIR.mkdir(parents=True, exist_ok=True)
     path = config_file()
@@ -90,6 +96,10 @@ def save_config(config: CliConfig) -> None:
         json.dumps(asdict(config), indent=2) + "\n",
         encoding="utf-8",
     )
+    # Filesystem may not support chmod (e.g. some FUSE mounts); don't fail
+    # the login flow over that.
+    with contextlib.suppress(OSError):
+        path.chmod(0o600)
 
 
 def get_api_url() -> str:
--- a/client/src/dhub/cli/init.py
+++ b/client/src/dhub/cli/init.py
@@ -3,9 +3,28 @@
 from pathlib import Path
 
 import typer
+import yaml
 from rich.console import Console
 
 console = Console()
+
+
+def _prompt_nonempty(message: str, *, max_length: int) -> str:
+    """Prompt repeatedly until the user enters a non-empty value of the right length.
+
+    The bare ``typer.prompt`` accepts an empty Enter, which then propagates
+    through to manifest validation as a confusing failure at publish time.
+    Catching it here keeps the round-trip tight.
+    """
+    while True:
+        value = typer.prompt(message).strip()
+        if not value:
+            console.print("[yellow]Value cannot be empty — please try again.[/]")
+            continue
+        if len(value) > max_length:
+            console.print(f"[yellow]Value must be at most {max_length} characters.[/]")
+            continue
+        return value
 
 
 def init_command(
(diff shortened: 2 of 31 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Broad superset sweep, but its themes are covered by the retained merge queue (#449, #448, #447, #446, #405, #438, #443, #375); its unique manifest caps are catalogued in #451. Branch preserved.

PR #378, opened 2026-05-23
refactor: harden subprocess, file perms, S3 timeouts, and DRY rate limiters

lfiaschi:
## Context

This PR is the output of a deep architecture-and-correctness review of the
codebase (server, CLI, shared, frontend). The review surfaced ~15 candidate
findings; this PR ships the ones that share a theme of **hardening
subprocess / network / secret handling without changing behaviour**, plus
one DRY refactor. Larger items (splitting the 3,465-line `database.py`,
adding a frontend error boundary, expanding the frontend test layer) are
called out at the bottom but deliberately deferred to focused PRs.

All 1,336 existing tests still pass; 12 new tests pin the new behaviour.
`ruff check`, `ruff format`, and `mypy` are clean.

### 1. Security: CLI token files were world-readable
`save_config()` in `client/src/dhub/cli/config.py` previously wrote the
user's API token via `Path.write_text`, inheriting the umask (typically
`0o644`). On a shared host (CI runner, multi-user dev box) any other
local user could read `~/.dhub/config.{env}.json` and obtain a token with
publish rights. Fixed by:

- Tightening `~/.dhub` to `0o700` on every save.
- Writing the token to a tmp file opened with `os.open(..., 0o600)` and
  atomically renaming it into place with `os.replace`, so the token is
  never on disk world-readable, not even briefly.
- Re-applying `chmod 0o600` after the rename to tighten any pre-existing
  loose permissions on the file.

### 2. Reliability: subprocess calls had no timeout
Three CLI code paths shelled out without a timeout, so a stalled remote
or wedged git index could hang the CLI indefinitely.

- `client/src/dhub/core/git_repo.py`: every `git` invocation now goes
  through a new `_run_git()` helper with a 5-minute timeout. `TimeoutExpired`
  is translated to a clean `RuntimeError` so callers handle a single
  exception type.
- `client/src/dhub/cli/registry.py`: `_detect_branch` got a 10-second
  timeout (the working tree is already on disk; this should finish in ms).
- `client/src/dhub/cli/runtime.py`: `uv sync` is now bounded at 10 minutes
  and `CalledProcessError` is caught so users see a friendly error rather
  than a raw traceback. `uv run` itself is intentionally unbounded —
  skills are arbitrary user code.

### 3. Reliability: S3 client had no explicit timeouts
`create_s3_client` in `server/src/decision_hub/infra/storage.py` accepted
boto3's defaults. We now build an explicit `botocore.config.Config` with
`connect_timeout=10s`, `read_timeout=60s`, and the modern `standard`
retry mode with `max_attempts=4`. A dead S3 endpoint will now surface a
clean error within seconds instead of stalling a Modal worker.

### 4. Code health: rate-limit boilerplate DRY
Three routes modules (`auth_routes.py`, `registry_routes.py`,
`search_routes.py`) contained eight near-identical 7-line
`_enforce_*_rate_limit` factory functions. Extracted a single
`make_rate_limit_dep(name, limit_attr, window_attr)` helper in
`rate_limit.py` and replaced all eight call sites. Net **~80 LOC removed**
with no behaviour change. Also dropped the now-unused `Request` import
from each module.

### 5. Correctness: rate-limiter stale-purge could fail to fire
`RateLimiter._purge_stale` was triggered by `total % 100 == 0`. Under
uneven traffic (e.g. several IPs cycling in and out below the modulo
threshold) the purge could never fire and the per-IP dict would grow
unbounded. Switched to a wall-clock cadence (`_PURGE_INTERVAL_SECONDS =
60`) which fires deterministically regardless of request volume.

## Verification

- `make test`: **941 server + 297 client + 98 shared = 1,336 tests pass**
- `make lint`: clean
- `make typecheck`: clean
- New tests cover every changed behaviour (12 new test cases across
  `test_rate_limit.py`, `test_storage_client.py`, `test_config.py`,
  `test_git_repo.py`).

## What was scoped out (deliberately)

These review findings warrant follow-up PRs rather than inflating this one:

| Finding | Why deferred |
|---|---|
| `infra/database.py` is a 3,465-line god module | Real architectural debt — natural seams exist for ~4 sub-modules (queries / skills+versions / trackers+audit / reporting) but it touches everything; needs its own PR. |
| Frontend has no error boundary / no error tracking | Adds a runtime dependency; UX decision needed. |
| Frontend has no request cancellation on unmount | Touches every `useApi` consumer; out of scope here. |
| Frontend test coverage thin (~8 tests) | Needs a real plan + Playwright setup. |
| TTLCache eviction prefers soonest-expiring entry | Suboptimal but correct; behaviour change would need benchmarks. |
| Anthropic API: no retry wrapper around `judge_eval_output` | Changes failure semantics for the eval pipeline; better as its own PR with explicit retry policy. |

## Audit findings the review got wrong (verified)

For the record, three flagged "bugs" turned out to be already-handled:

- `find_audit_logs` **already** has `ORDER BY created_at DESC, id DESC`
  with a unique tiebreaker (`database.py:2480`).
- `judge_eval_output` **already** passes `timeout=60` (`anthropic_client.py:68`).
- Publish pipeline **already** rolls back the DB version + audit on S3
  upload failure (`publish_pipeline.py:754-768`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01AoXmxTfgcaReSCHQy6Zz6s)_


lfiaschi:

PR #379, opened 2026-05-25
refactor: rate-limit hygiene, atomic CLI config, single-query summary, mobile font

lfiaschi:
## Summary

Six self-contained fixes surfaced by an architectural review pass over the repo. No public API or response-shape changes; all 1,431 tests stay green (953 server + 298 client + 98 shared + 82 frontend), ruff + format + mypy clean.

The review and refactor were done in a two-step workflow: a system map + finding catalog, then this focused PR addressing the top high-leverage items that are low risk and small enough to land together.

### 1. Rate-limit DRY + proxy-aware client IP — `server/`

`registry_routes.py` had **seven near-identical 9-line `_enforce_*_rate_limit` helpers**. Replaced with a single factory that takes the bucket name plus the two `Settings` attribute names. Adding a new limited endpoint is now a one-liner.

`rate_limit.py` had two real bugs in the same file:

- **`request.client.host` is the proxy address** when the app runs behind Modal / Cloud Run / nginx, collapsing every public client into a single bucket and rendering the per-IP limiter useless. Added a `client_ip()` helper that honours `X-Forwarded-For` (leftmost = original client) and `X-Real-IP` *only* when the new `trust_proxy_headers` setting opts in. Default stays `False` so we don't trust user-controllable headers without explicit operator wiring.
- **The memory-bound trick (`if total % 100 == 0: prune`) was broken.** `total = sum(len(v) for v in self._requests.values())` was O(num_keys) on every request, the trigger fired on *every* call under sparse traffic (most keys at count 1) and *almost never* under load, and per-key buckets could grow unbounded within a window. Replaced with: unconditional per-key pruning, a wall-clock full sweep every 60s, and a hard cap of 50 000 tracked keys with oldest-touched eviction.

### 2. Atomic CLI config write + 0o600 perms — `client/`

`~/.dhub/config.*.json` holds a long-lived bearer token. The old `Path.write_text()` was **non-atomic** (a crash mid-write would truncate the file, forcing the user to delete and re-authenticate — the existing code at `config.py:67-73` even has a user-facing error message for this exact case) and the **default umask of 0o022 left the file world-readable** on shared systems.

Rewrote `save_config()` to:
- write to a sibling tempfile via `mkstemp` (same dir = same FS = atomic `os.replace`),
- `fsync` before rename so a power loss doesn't lose the write,
- explicitly `chmod 0o600` on the tempfile *and* on the destination (covers Windows ACL quirks and existing loose-perm files left by older CLI versions),
- clean up the tempfile on any error so the original is never touched.

Added `is_config_file_secure()` for `dhub doctor` to surface loose perms.

### 3. Single-query skill-summary endpoint — `server/`

`GET /v1/skills/{org}/{name}/summary` was making **four sequential DB calls** per request — `find_skill_by_slug` + `resolve_latest_version` + `find_org_by_slug` + `has_active_tracker_for_repo` — even though every field they returned is already on the denormalised skills row used by the list endpoint.

Added `fetch_skill_summary_by_slug()` which reuses `_SKILL_SUMMARY_COLUMNS` and the same correlated `tracker_exists` subquery the list endpoint already uses. One SELECT, identical visibility filtering, no JOIN duplication. The handler now just maps row → response.

### 4. Responsive body font-size — `frontend/`

`body { font-size: 18px }` on `index.css:66` violated the project's mobile-first rule from `CLAUDE.md`. Every `--text-*` variable is in `rem`, so the fixed 18px anchor pushed the 4rem hero off the right edge of a 375px iPhone. Changed to `clamp(15px, 1rem + 0.4vw, 18px)` — floors at 15px on narrow phones and ceilings at 18px from ~720px up, preserving the desktop look.

### 5. Settings

New `trust_proxy_headers: bool = False` to opt into proxy-aware client-IP detection. Documented to warn that it must never be enabled when the app is directly internet-exposed.

### 6. Tests (+27 new cases)

| File | Added | Covers |
|---|---|---|
| `test_rate_limit.py` | +9 cases | `client_ip` proxy-header behaviour (5), per-key prune bound, time-based full prune, key-cap eviction, proxy-aware bucketing |
| `test_config.py` | +7 cases | `TestSaveConfigSecurity`: mode 0o600, loose perms tightened, mid-write crash leaves original byte-for-byte intact, no tempfile leaks, dir mode. `TestIsConfigFileSecure`: owner-only, world-readable detection, missing file |
| `test_registry_routes.py` | +5 cases | New `TestGetSkillSummary` — the endpoint had **zero server-side tests**. Happy path, 404, auth→org-id wiring, anonymous path, nullable fields |
| `test_seo_routes.py` (new) | +6 cases | `seo_routes.py` had **zero coverage**. Static page inclusion, skill/org rows, XML escaping of user-supplied slugs, cache short-circuit, prod-vs-non-prod robots |

## Out of scope (deferred from review)

The review surfaced several items I chose not to bundle here to keep the PR reviewable:

- **JWT TTL** (1 year + embedded `github_orgs`) — policy decision, needs product input.
- **Frontend test coverage gap** (3 test files vs 29 components, no e2e) — multi-week scope; recommend a separate Playwright sprint.
- **No `/metrics` endpoint** — infra change, needs a chosen exporter (prometheus_client vs OTLP).
- **`runtime.py` subprocess calls without timeout** — needs UX decision on `uv sync` and user-skill timeouts.
- **`deps.py:78` invalid-JWT IP logging at WARN** — minor, defer.

## Test plan

- [x] `make lint` clean (ruff check + format)
- [x] `make typecheck` clean (mypy across 88 files)
- [x] `make test-server` 953 passed
- [x] `make test-client` 298 passed
- [x] `make test-shared` 98 passed
- [x] `make test-frontend` 82 passed
- [x] `npx tsc -b` clean (frontend)
- [ ] CI green on PR
- [ ] Manual verify: hit `/v1/skills/<org>/<skill>/summary` on dev after deploy
- [ ] Manual verify: rate limiter no longer collapses to one bucket on dev (set `trust_proxy_headers=true` in `.env.dev` first)

https://claude.ai/code/session_01G1C6iQK3pcL3zv4GrN73xA

---
_Generated by [Claude Code](https://claude.ai/code/session_01G1C6iQK3pcL3zv4GrN73xA)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its 4-query→1 skill summary and is_config_file_secure() are catalogued in #451. Branch preserved.

PR #381, opened 2026-05-26
refactor(api): DRY rate-limit deps, fix purge cadence, drop redundant eval-run read

lfiaschi:
## Summary

Three focused improvements to the FastAPI layer, surfaced by an architect + principal-engineer codebase review. Behaviour is unchanged for clients — these are pure refactors / latency fixes / dead-DB-trip removals.

| # | Change | Category | Risk |
|---|---|---|---|
| 1 | DRY 10 duplicated `_enforce_*_rate_limit` helpers behind one factory | code health | low |
| 2 | Fix `RateLimiter._purge_stale` cadence bug (stale buckets could leak indefinitely) | bug fix | low |
| 3 | Drop the redundant DB re-read in `GET /v1/eval-runs/{run_id}` zombie path | performance | low |

Net diff: **+212 / -139 lines**, with the source code shrinking and the tests growing.

---

## 1. DRY: collapse 10 rate-limit dependency helpers into one factory

`registry_routes.py`, `auth_routes.py`, and `search_routes.py` each contained near-identical helpers like:

```python
def _enforce_publish_rate_limit(request: Request) -> None:
    """Rate-limit the publish endpoint."""
    state = request.app.state
    if not hasattr(state, "_publish_rate_limiter"):
        state._publish_rate_limiter = RateLimiter(
            max_requests=settings.publish_rate_limit,
            window_seconds=settings.publish_rate_window,
        )
    state._publish_rate_limiter(request)
```

Ten copies of this pattern, differing only in the settings prefix. Replaced with one factory in `api/rate_limit.py`:

```python
```

`make_rate_limit_dep(name)` reads `settings.{name}_rate_limit` / `settings.{name}_rate_window` lazily on first call and caches the limiter under `app.state.rate_limiters[name]` (a single dict instead of N per-name attributes).

The returned callable carries a descriptive `__name__` so it still shows up sensibly in FastAPI tracebacks / OpenAPI errors.

## 2. Bug fix: `RateLimiter._purge_stale` cadence

**Before**: periodic eviction was gated on

```python
if total % 100 == 0:
    self._purge_stale(cutoff)
```

`total` reflects *instantaneous* bucket sizes, not a request counter. At steady-state (e.g. ~99 active timestamps across all IPs with steady churn) it can hover just below the modulo boundary indefinitely, leaking idle IP buckets and growing memory unbounded relative to the active traffic.

**After**: an explicit `_checks_since_purge` counter triggers exactly once per `_PURGE_EVERY_N_REQUESTS` calls — the cadence the original comment ("Check every 100 requests") had always claimed.

Regression test `test_purges_stale_ips_after_threshold_requests`:

1. Stamps 50 distinct IPs with one timestamp each.
2. Advances the monotonic clock past the window.
3. Drives a single hot IP enough times to trigger one purge.
4. Asserts all 50 idle IPs are gone and the hot IP survives.

## 3. Perf: drop redundant DB re-read in `get_eval_run`

The handler used to read the run, call `_check_zombie` (which may UPDATE the row), then re-read the run unconditionally:

```python
run = find_eval_run(conn, parsed_id)
...
_check_zombie(conn, run)
# Re-read after potential zombie update
run = find_eval_run(conn, parsed_id)
return _run_to_response(run)
```

That second read fires on **every poll**, even when the run was not stale — the common case. The frontend polls this endpoint while an eval is running, so this was a measurable wasted round-trip.

`_check_zombie` now mirrors the DB write onto the in-memory frozen dataclass (`object.__setattr__(run, "status", "failed")`, etc.) so the caller's local `run` reflects what a re-read would have returned. The handler skips the second `find_eval_run` call.

The existing `test_zombie_detection_marks_stale_run_as_failed` was rewritten to assert `mock_find_run.call_count == 1` as a regression guard. The mutated dataclass still serializes correctly through the Pydantic response model (verified by the existing assertion on `resp.json()["status"] == "failed"`).

## What I considered and rejected for this PR

The full review surfaced ~13 high-leverage findings; I kept this PR scoped to three safe, mechanical fixes. Deliberately deferred to follow-ups:

- Splitting `database.py` (3.5 kLoC) and `registry_routes.py` (1.4 kLoC) by resource — mechanical but very large diffs.
- Replacing 5 `_build_*_fn` Gemini factories in `publish_pipeline.py` with one helper — touches the safety pipeline; needs deeper review.
- Switching `download_skill` from a server-side proxy to a 302 to a presigned S3 URL — behavioural change for clients.
- Caching the GitHub App installation token in `_resolve_github_token` (currently re-minted per Modal container) — straightforward but Modal-runtime-sensitive; worth its own PR.
- Adding CSP to `SecurityHeadersMiddleware`.
- Fixing the misleading `check_org_membership` docstring + 302 status handling.

## Test plan

- [x] `pytest server/tests/test_api/test_rate_limit.py` — 10 passed (was 5; 5 added).
- [x] `pytest server/tests/test_api/test_eval_logs.py` — 21 passed (1 updated to assert single-read invariant).
- [x] `pytest server/tests/test_api/test_registry_routes.py test_search_routes.py test_auth_routes.py` — 86 passed, no behaviour regressions.
- [x] Full server suite: `pytest -m "not slow"` — **938 passed**.
- [x] Shared suite: 98 passed.
- [x] Client suite: 291 passed, 29 skipped.
- [x] `ruff check` clean.
- [x] `ruff format --check` clean.
- [x] `mypy` clean on touched files.

The `uv.lock` change is the workspace lockfile catching up to the already-committed `pyproject.toml` versions (`dhub-cli 0.12.1 → 0.12.3`, `dhub-core 0.2.8 → 0.2.10`); not introduced by this PR but unavoidable when running `uv` against the current branch.

https://claude.ai/code/session_01A5ysN465cjLw4RPsjE43Zc

---
_Generated by [Claude Code](https://claude.ai/code/session_01A5ysN465cjLw4RPsjE43Zc)_


lfiaschi:

PR #383, opened 2026-05-29
fix: review-driven correctness, security, and convention hardening

lfiaschi:
## Summary

A focused, low-risk batch of fixes surfaced by a deep architect/principal-engineer review of the codebase. Every behavioral change ships with a regression test. Deliberately **excludes** the contested large refactors (proxy-aware rate-limit keying, the rate-limit-factory consolidation) so this PR stays small and safe to merge.

All quality gates pass locally: `ruff check` + `format`, `mypy` (88 files), and the full test suites — **shared 36, client 297, server 935, frontend 82**.

### Security
- **`decode_jwt` now requires `exp` + `sub`** (`domain/auth.py`). python-jose ignores PyJWT's `require` list, so this uses `require_exp`/`require_sub` — a token minted without an expiry can no longer be accepted as valid forever.
- **CLI config written `0600` (dir `0700`)** (`cli/config.py`). The config holds the bearer token; previously it inherited the default umask (typically world-readable), exposing credentials to other local users on a shared host or CI runner.

### Correctness / bugs
- **Gemini stops retrying `403`** (`infra/gemini.py`). 403 is a hard auth/permission failure (e.g. a bad `GOOGLE_API_KEY`), not transient — it now fails fast instead of burning the gauntlet's retry budget. Google signals rate limits with `429`, which stays retriable.
- **`fetch_skills_by_repo` `.git/` normalization** (`infra/database.py`). The SQL now mirrors `_normalize_repo_url`'s order (strip trailing slash → strip `.git`). Previously `https://github.com/o/r.git/` normalized to `r` in Python but `r.git` in SQL, hiding the skill from the repo page.
- **Search determinism** (`infra/database.py`). `search_skills_hybrid` FTS/vector queries get a unique `id` tiebreaker so rank/distance ties produce a deterministic page and a stable vector-vs-FTS dedup winner (per the project's "every LIMIT needs a unique ORDER BY tiebreaker" rule).
- **`parse_skill_ref` rejects empty halves** (`core/validation.py`). `"org/"`, `"/skill"`, and `"/"` previously parsed as valid and could resolve to the org directory itself (e.g. `~/.dhub/skills/org/`).
- **`get_eval_run` concurrent-delete guard** (`api/registry_routes.py`). Under READ COMMITTED the re-read after zombie-check could return `None`; it now returns 404 instead of raising on a `None` deref.
- **`find_audit_logs` offset** (`infra/database.py`). OFFSET is only applied when bounded by a LIMIT, so an `limit=None` ("all rows") caller no longer silently gets the unbounded tail past the offset.
- **`parse_frontmatter_yaml` honors its `-> dict` contract** (`dhub_core/manifest.py`). Empty/comment-only frontmatter (`yaml.safe_load` → `None`) and bare scalars now collapse to `{}`.
- **`CLIVersionMiddleware` latent unbound-name** (`api/app.py`). `client_parsed` is now initialized on all paths, removing a trap that a future reorder of the short-circuit could turn into a `NameError` on every `/v1/` request.

### Code health (simplify)
- Consolidated the two duplicated `_parse_uuid` helpers into a single `deps.parse_uuid`; removed a doubled `# noqa` and the now-unused `UUID` imports.

### Frontend conventions
- Replaced hardcoded `rem` font sizes in `GradeBadge.module.css` and `OrgDetailPage.tsx` with the `--text-*` scale variables (per the fixed type-scale convention).

### Tests added
JWT claim requirements (missing `exp`/`sub` rejected), Gemini 403 fast-fail, `parse_skill_ref` rejection cases, frontmatter dict contract (empty/comment/scalar), and CLI config file/dir permissions.

## Test plan
- [x] `ruff check` + `ruff format --check` clean
- [x] `mypy client/src server/src shared/src` — no issues (88 files)
- [x] `pytest shared/tests` — 36 passed
- [x] `pytest client/tests` — 297 passed
- [x] `pytest server/tests -m "not slow"` — 935 passed
- [x] frontend `tsc -b` + `eslint` clean; `vitest run` — 82 passed

> Context: this branch is one of several review-driven runs; it intentionally scopes to the safe, high-confidence subset and leaves the proxy-IP rate-limiting and rate-limit-factory work (which need infra verification / are larger) as follow-ups.

https://claude.ai/code/session_01MoX46viFKLatUfrGBXUfqi

---
_Generated by [Claude Code](https://claude.ai/code/session_01MoX46viFKLatUfrGBXUfqi)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -82,14 +82,20 @@ def load_config() -> CliConfig:
 def save_config(config: CliConfig) -> None:
     """Save CLI config to ~/.dhub/config.{env}.json.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ~/.dhub directory if it does not already exist. The config
+    file holds the bearer token, so it is written with owner-only (0600)
+    permissions — and the directory 0700 — to prevent other local users on a
+    shared host or CI runner from reading the credentials.
     """
-    CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+    CONFIG_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
+    # mkdir's mode is ignored when the dir already exists, so enforce it.
+    os.chmod(CONFIG_DIR, 0o700)
     path = config_file()
     path.write_text(
         json.dumps(asdict(config), indent=2) + "\n",
         encoding="utf-8",
     )
+    os.chmod(path, 0o600)
 
 
 def get_api_url() -> str:
--- a/client/src/dhub/core/validation.py
+++ b/client/src/dhub/core/validation.py
@@ -20,9 +20,13 @@ def parse_skill_ref(skill_ref: str) -> tuple[str, str]:
     """Parse 'org/skill' reference into (org_slug, skill_name).
 
     Raises:
-        ValueError: If the reference is not in org/skill format.
+        ValueError: If the reference is not in org/skill format, or if either
+            the org or skill component is empty (e.g. 'org/' or '/skill').
     """
     parts = skill_ref.split("/", 1)
-    if len(parts) != 2:
+    # ``"org/".split("/", 1)`` yields ``["org", ""]`` — length 2 but with an
+    # empty half — so guard the components explicitly. An empty skill name
+    # would otherwise resolve to the org directory itself (e.g. ~/.dhub/skills/org/).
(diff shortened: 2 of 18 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its JWT exp/sub validation and .git/ URL-normalization fixes are catalogued in #451. Branch preserved.

PR #385, opened 2026-05-31
fix: HTTP timeouts, token leak in clone_repo, stale tab on skill nav

lfiaschi:
## Summary

A focused bundle of correctness / security / DX fixes surfaced by a deep architectural review of the codebase. Each change is independently revertable and ships with a new test.

The review identified ~12 high-leverage issues across server, client and frontend. This PR addresses the five lowest-risk, highest-value ones; the structural items (splitting the 3,465-line `database.py`, fixing the commit-before-S3-upload race in `publish_pipeline.py`, extracting a shared CLI HTTP client, etc.) are intentionally left to dedicated follow-up PRs so reviews stay tractable.

### Server

**`infra/github.py` — add timeouts to every `httpx.AsyncClient()`**
All 7 call sites previously used httpx's default of *no timeout*. A stalled GitHub API call (slow upstream, DNS blackhole, slowloris) could pin a FastAPI worker indefinitely and exhaust the pool. Now each call carries a shared `httpx.Timeout(15.0)` constant. Concrete failing scenario before this fix: if `api.github.com` becomes unreachable mid-request, `check_org_membership()` (called on every authenticated login) blocks forever and starves the event loop.

**`domain/repo_utils.py` — sanitize stdout, not just stderr, on clone failure**
`clone_repo()` rewrites the remote URL to embed a GitHub token (`https://x-access-token:<TOKEN>@github.com/...`). On failure, only `stderr` was scrubbed before being attached to the `CalledProcessError`. Git can echo the same URL to `stdout` depending on the failure path, and that stream was kept raw. Result: the token could leak into downstream logs, crash reporters, or audit pipelines. Both streams are now scrubbed.

### Frontend

**`pages/SkillDetailPage.tsx` — reset `activeTab` when navigating between skills**
The state-reset effect (line 73) covers zip/files/markdown state but `activeTab` was missing. Navigating from skill A's "Audit Log" tab to skill B kept "Audit Log" selected on first paint of skill B, even though that tab is unrelated to the new skill. This violates the CLAUDE.md rule *"Reset state on context changes"*. Now reset together with the other state.

**`pages/SkillDetailPage.tsx` + `components/ScannerReport.tsx` — stable list keys**
Four list renders used bare `key={i}`:
- `CheckResultsGrid` (audit check results, which carry a per-row expanded sub-state)
- `llm_primary_threats`, `actionableCorrelations`, `meta_recommendations` in ScannerReport

Each can reorder, filter (false-positives are toggled), or partially refresh, so index keys can corrupt the highlighted/expanded sub-state. Switched to composite keys built from a natural identifier plus index (`${check_name}-${i}`, etc.), which keeps reconciliation correct without assuming names are globally unique.

### Tests

- `test_infra/test_github.py::TestGithubTimeouts` — asserts the module timeout is finite *and* that `AsyncClient` actually receives it (verified by patching `httpx.AsyncClient` and inspecting `__init__` kwargs).
- `test_scripts/test_github_crawler.py::TestCloneRepo::test_token_sanitized_from_stderr_and_stdout` — simulates git echoing the authenticated URL to both streams and asserts the token is replaced with `***` in both `output` and `stderr` of the raised `CalledProcessError`.
- `pages/SkillDetailPage.test.tsx::resets activeTab to overview when navigating to a different skill` — exercises real `useNavigate` (not a remount) so the URL change runs through the same effect path users would hit.

## Test plan

- [x] `pytest server/tests/test_infra/test_github.py -v` — 15/15 pass (incl. 2 new)
- [x] `pytest server/tests/test_scripts/test_github_crawler.py -v` — 92/92 pass (incl. 1 new)
- [x] `pytest server/tests/test_infra/ server/tests/test_domain/test_tracker_service.py -m "not slow"` — 228 pass / 24 deselected
- [x] `npx vitest run` (frontend) — 83/83 pass (incl. 1 new)
- [x] `npx tsc -b` + `npx eslint` — no new warnings (one pre-existing warning on `addRecentlyViewed` effect, unrelated)

## Not in this PR (follow-ups)

From the same review — left out to keep this diff reviewable:

1. **`publish_pipeline.py:750–768`**: DB commits *before* S3 upload, so a concurrent reader between commit and rollback sees a version with no zip. Should be upload-then-commit (or use a soft-delete state).
2. **`tracker_service.py:517–571`**: Modal `fn.map()` deadline early-exit never appends to `completed_tracker_ids`, so `_persist_orphaned_tracker_errors` overwrites every deferred tracker's `last_error` with a generic message.
3. **`client/cli/registry.py`**: 22 hand-written `httpx.Client(timeout=60)` blocks — extract a shared `dhub.core.api` client with auth, timeout, and JSON-safety baked in.
4. **`infra/database.py`** (3,465 LOC): split by domain (skills/orgs/trackers/evals/audit) into a `queries/` subpackage.
5. Frontend test coverage for `SkillsPage`, `OrgDetailPage`, `EvalReportView`, `ScannerReport` (none of these have tests today despite being among the largest components).

https://claude.ai/code/session_01Ap3m3DqdhTSDXyNRtvzJLc

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ap3m3DqdhTSDXyNRtvzJLc)_


lfiaschi:

PR #386, opened 2026-06-02
fix(api): rate-limit per real client IP, factor out limiter setup, escape LIKE wildcards

lfiaschi:
## Why

Three small, related correctness fixes plus a refactor that the others surfaced. All shipped behind existing tests — no behavioural surprises for callers.

### 1. Per-IP rate limiting now uses the real client IP

`RateLimiter` keyed on `request.client.host`, which behind Modal's edge is the load-balancer's IP, not the user's. In effect every public reader shared a single bucket: one heavy user could 429 everyone else, and a single attacker could not be isolated.

A new `client_ip_from_request()` helper:
1. Picks the first hop in `X-Forwarded-For` (what Modal / CloudFront set).
2. Falls back to the socket peer when no XFF (local dev / direct connection).
3. Falls back to `"unknown"` so per-request never crashes.

Applied in `RateLimiter` and in `deps.get_current_user` (for its "Invalid JWT from {}" log line, which had the same bug for forensic logging).

### 2. Single `make_rate_limit_dep("<name>")` replaces 10 copies

`api/registry_routes.py` (lines 86–167), `api/auth_routes.py` and `api/search_routes.py` each shipped near-identical `_enforce_*_rate_limit` functions — 10 in total — doing the same lazy `hasattr` init dance. Each copy also had a small benign race: two early requests could both pass `hasattr` and the second's `RateLimiter` would clobber the first, resetting any IPs already tracked.

`make_rate_limit_dep("widget")`:
- Reads `settings.widget_rate_limit` and `settings.widget_rate_window`.
- Lazy-installs on `app.state` under `_widget_rate_limiter`.
- Init is guarded by a module-level lock with double-checked read.
- Returns a dep whose `__name__` matches the old helper, so tracebacks stay greppable.

Net: ~80 lines removed across three route files; one place to evolve the pattern.

### 3. Periodic purge of stale IPs is now deterministic

The old `RateLimiter._purge_stale` fired only when `total % 100 == 0`, where `total` was the sum of all currently active timestamps across every IP. That number can skip 0 entirely as old entries expire — leaving stale IPs in memory indefinitely in a long-lived container. Replaced with a per-instance call counter (`_PURGE_EVERY_N_CALLS = 256`) that triggers reliably.

### 4. `LIKE` escape in two batch update queries

`batch_update_github_stars` (database.py:1197) and `batch_update_github_repo_metadata` (database.py:1220) interpolated repo URLs into a LIKE pattern with no escape. Repo URLs commonly contain `_`, which is a single-char LIKE wildcard:

- Before: a star refresh for `github.com/foo_bar/baz` would also UPDATE the skill at `github.com/fooXbar/baz`, silently overwriting its star count.
- After: uses the existing `_escape_like` helper with `escape="\\"`, matching the pattern already used by the tracker queries at lines 3135 / 3163.

The other LIKE call sites in the file are already correct — this fix brings these two stragglers in line.

## Tests

**29 new test cases** across the two test files, all passing:

* `tests/test_api/test_rate_limit.py`
  * Full coverage of `client_ip_from_request`: XFF priority, whitespace handling, empty/comma-only headers, no-client edge case.
  * `test_real_client_ip_isolated_behind_proxy`: regression case proving two users behind one LB get separate buckets.
  * `test_periodic_purge_removes_stale_ips`: shrinks the heartbeat to 5 and verifies a stale IP is dropped after 5 calls from a different IP.
  * `TestMakeRateLimitDep`: lazy init, idempotent reuse, 429 enforcement, dep name preservation.
* `tests/test_infra/test_database_by_repo.py`
  * Renders the SQL with `compile(compile_kwargs={"literal_binds": True})` and asserts the escaped `_` / `%` and the trailing `ESCAPE '\'` appear in the statement — for both `batch_update_github_stars` and `batch_update_github_repo_metadata`.

## Verification

```bash
$ DHUB_ENV=dev uv run --package decision-hub-server --extra dev pytest tests/ -m "not slow"
948 passed, 34 deselected in 28.69s
$ uv run --package dhub-cli --extra dev pytest client/tests/
291 passed, 29 skipped in 8.68s
$ uv run --package dhub-core --extra dev pytest shared/tests/
98 passed in 0.17s
$ uvx ruff@0.15.3 check .                              # All checks passed!
$ uvx ruff@0.15.3 format --check .                     # 197 files already formatted
$ uvx mypy client/src/ server/src/ shared/src/         # Success: no issues found in 88 source files
```

## Scope I intentionally did not touch in this PR

The codebase review surfaced several larger items worth their own PRs:
- Decomposing `database.py` (3,465 lines) into per-entity query modules behind the existing public API.
- Splitting `registry_routes.py` (1,352 lines) into sub-routers and a `schemas/` module.
- Adding direct unit tests for `publish_pipeline.py` (848 lines, currently zero direct coverage).
- Sanitising user-controlled filenames before they hit LLM safety judges in `gauntlet.py` / `gemini.py` (prompt-injection vector).
- Shortening the 1-year JWT lifetime and adding a revocation list.

Happy to follow up on any of these if you want — kept this PR small so it's easy to review and easy to roll back.

## Test plan
- [x] `make test-server` (`pytest -m "not slow"`) — 948 pass
- [x] `make test-client` — 291 pass (29 pre-existing skips unrelated)
- [x] `make test-shared` — 98 pass
- [x] `make lint` — ruff check + format clean
- [x] `make typecheck` — mypy clean
- [ ] CI on this PR

https://claude.ai/code/session_01Ms3M9WP6GdDmK2z8MQQ57G

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ms3M9WP6GdDmK2z8MQQ57G)_


lfiaschi:

PR #389, opened 2026-06-05
refactor: DRY rate limiters, Gemini callbacks, and CLI HTTP client

lfiaschi:
## Why

I ran an architecture + code-health pass across the monorepo (`server/api/`, `server/domain/`, `server/infra/`, `client/cli/`, `frontend/`, all test suites). Three duplications kept showing up and together account for >250 lines of boilerplate spread across 13 files. They are all the same shape — a small piece of plumbing that grew copies as endpoints / commands were added — so a single factory eliminates each one cleanly without any behaviour change.

This PR is a pure refactor: existing 933 server + 291 CLI + 98 shared tests still pass, plus **36 new unit tests** lock in the contracts of the new helpers.

## Findings being addressed

| # | Duplication | Files | Lines removed |
|---|-------------|-------|---|
| 1 | 9 `_enforce_<name>_rate_limit` boilerplates | `api/registry_routes.py`, `api/search_routes.py`, `api/auth_routes.py` | ~85 |
| 2 | 5 `_build_*_fn` Gemini callback builders | `domain/publish_pipeline.py` | ~75 |
| 3 | 38 inline `httpx.Client(timeout=60)` + manual headers in CLI | `cli/*.py` (4 of 9 migrated here) | ~30 (partial) |

### 1. `server/api/rate_limit.py` — `make_rate_limit_dep()` factory

Each route module used to declare its own `_enforce_<name>_rate_limit(request)` function that lazily built a `RateLimiter`, cached it on `request.app.state`, and delegated. The pattern was copy-pasted 9 times with only the names and Settings attribute lookups changing.

New factory:

```python
_enforce_publish_rate_limit = make_rate_limit_dep(
    "publish", "publish_rate_limit", "publish_rate_window"
)
```

The factory caches the limiter on `app.state` under `_{name}_rate_limiter` (matching the existing attribute names, so any tooling that pokes at app state keeps working), sets `__name__` for readable FastAPI traces, and reads max/window from named Settings attributes.

### 2. `server/domain/publish_pipeline.py` — `_make_gemini_callback()`

The 5 `_build_*_fn` builders all followed the same recipe: short-circuit when `google_api_key` is empty, lazily import a function from `decision_hub.infra.gemini`, share an httpx client, and return a closure that forwards positional args with `model=settings.gemini_model`.

Extracted into a single `_make_gemini_callback(settings, fn_name, gemini)` that resolves the target via `getattr(gemini_module, fn_name)`. The 5 named wrappers are **intentionally preserved** as 1-line pass-throughs because existing tests `@patch("...publish_pipeline._build_analyze_fn", ...)` rely on them as monkey-patch handles (see `tests/test_api/test_registry_routes.py` and `tests/test_api/conftest.py`). Future changes to the factory body will not touch those tests.

### 3. `client/cli/config.py` — `api_client()` context manager

Added a `DEFAULT_HTTP_TIMEOUT = 60.0` constant and a context-manager helper that yields a pre-configured `httpx.Client`:

```python
with api_client(token=get_token()) as client:
    resp = client.post("/v1/keys", json={...})
    raise_for_status(resp)
```

`base_url` is set to `get_api_url()`, default headers carry the CLI version + optional bearer, and the timeout defaults to 60s (Modal cold-start floor). The 38 inline `httpx.Client(timeout=60)` sites are not all migrated here — only the 4 small command modules (`keys`, `org`, `access`, `search`) are converted as a reference pattern. The large `registry.py` (~25 sites) is left for a follow-up so this PR stays reviewable.

A note on import style: imports of `get_token` / `get_optional_token` / `api_client` are kept inside command functions because the existing test suite patches at `dhub.cli.config.<name>`, which only works if commands rebind on each call. This convention is documented inline.

## Tests added (36 total, all in this PR)

| File | New tests | What they lock in |
|---|---|---|
| `server/tests/test_api/test_rate_limit.py` | +5 | lazy construction reads named Settings attrs; cached limiter is reused across calls; budget is enforced through the cache; same-Settings deps still get isolated state; `__name__` is meaningful for traces |
| `server/tests/test_domain/test_publish_pipeline_callbacks.py` (new) | +10 | factory short-circuits without an API key; supplied Gemini client is reused (no double-create); closure forwards positional args + `model=`; each of the 5 wrappers routes through the factory |
| `client/tests/test_cli/test_config.py` | +5 | `api_client()` sets `base_url`, adds CLI version header, attaches bearer when given, omits it when `None`, honours custom + default timeouts |

## Diff stats

```
14 files changed, 460 insertions(+), 274 deletions(-)
```

Net +186 — and ~200 of the additions are new tests, so non-test source is net **negative**.

## Quality gates

```
$ uv run --package decision-hub-server --extra dev pytest server/tests/ -m "not slow"
948 passed, 34 deselected in 23.07s

$ uv run --package dhub-cli --extra dev pytest client/tests/
296 passed, 29 skipped in 6.10s

$ uv run --package dhub-core --extra dev pytest shared/tests/
98 passed in 0.12s

$ uvx ruff@0.15.3 check .            → All checks passed!
$ uvx ruff@0.15.3 format --check .   → 198 files already formatted
$ uvx mypy <changed files>           → no issues found in 10 source files
```

## Other findings flagged for follow-up (not in this PR)

The review surfaced higher-impact but riskier items. Calling them out so the team can prioritise; each one wants its own focused PR:

- **`infra/anthropic_client.py`** has no retry — a single 429/503 from Anthropic fails the entire eval pipeline. Compare with `infra/gemini.py`'s 3-retry exponential backoff.
- **`modal_app.py`** mints one GitHub App installation token at the start of the nightly crawl, then shares it across every downstream container. Token rotation mid-crawl is impossible. Each container should mint its own (which `tracker_process_repo` already does).
- **`domain/gauntlet.py`** (1,355 lines) bundles regex, taint tracking, entropy scoring, and LLM judges. Splitting into `gauntlet/{regex,taint,entropy,llm_judges}.py` would make each layer independently testable.
- **`infra/database.py`** (3,465 lines) keeps growing. Denormalised `latest_*` columns on `skills` are refreshed manually with no rollback guard if `_refresh_skill_latest_version()` fails mid-transaction.
- **Frontend/server type drift** — `frontend/src/types/api.ts` is hand-mirrored from the server's Pydantic models with no contract test guarding both sides. Adding a removed/renamed field silently breaks the SPA at runtime.
- **Zero end-to-end tests** — no Playwright/Cypress. The test pyramid is heavily bottom-loaded with mock-driven unit tests.

https://claude.ai/code/session_01C3aCYYbZ9NPHQCDaPgATiL

---
_Generated by [Claude Code](https://claude.ai/code/session_01C3aCYYbZ9NPHQCDaPgATiL)_


lfiaschi:

PR #392, opened 2026-06-08
refactor(rate-limit,cache,ziputil): proxy-aware limits, safer eviction, symlink check

lfiaschi:
## Summary

Surfaced during a deep architect/principal-engineer review of the repo. These are the highest-leverage, lowest-risk fixes from that pass — purely internal changes (no API contract, no env defaults change without an explicit opt-in) and they ship with regression tests.

The full review identified ~13 items; this PR ships the focused subset that's self-contained and safe to land independently. The remaining items (JWT lifetime, splitting `execute_publish`, sharing the ask ThreadPool, GitHub org re-validation) are best as separate PRs because they touch product surface or larger code areas.

### What changed and why

**1. Rate limiter is now proxy-aware.** `request.client.host` is the Modal edge IP, not the caller's. Without `X-Forwarded-For` awareness, every public client collapsed into one bucket. Opt-in via new `trust_proxy_hops` setting (default `0` — zero behaviour change unless you set it). The leftmost-of-rightmost-N XFF entry is the IP added by the last trusted proxy, so an untrusted hop can't spoof which client we bucket on. Tests cover single-hop, multi-hop, hop-count overflow, and empty/malformed XFF.

**2. Rate limiter purges on time, not request count.** The old code only purged when `total % 100 == 0`, and even then only dropped IPs whose latest timestamp was already older than the cutoff. Under unique-IP fan-out (a many-source DoS) nothing was ever stale enough to evict and memory grew unbounded between organic timeouts. Now we record `_last_purge_at` and purge once per `window_seconds` regardless of traffic shape. Regression test exercises 500 distinct IPs in one window and asserts the bucket dict shrinks back to active clients after the window passes.

**3. Rate-limit dependency factory.** `rate_limit_dep(state_attr, max_attr, window_attr)` replaces 9 nearly-identical `_enforce_*_rate_limit` closures across `registry_routes.py`, `search_routes.py`, and `auth_routes.py`. Drops ~85 lines of boilerplate and three duplicated patterns that were already drifting (different attribute-presence checks, different settings access styles).

**4. TTLCache eviction no longer mutates during iteration.** The previous `_evict_one` did `del self._store[k]` inside a `for k, entry in self._store.items()` loop, only avoiding the `RuntimeError: dictionary changed size during iteration` because of an early `return`. The new version picks the eviction target first then mutates once. Added a regression test that hammers `set` past `max_size` repeatedly (forces "no expired entries" branch every time) and a second test asserting the soonest-to-expire entry is dropped when nothing is expired.

**5. ziputil follows symlinks before the prefix check.** `os.path.normpath` only collapses `..` lexically — a symlink inside the target dir could let an extracted file escape on disk even though the textual path looked safe. Switched both target and resolved entry to `os.path.realpath`. Two new tests: a symlinked subdir attempting to escape (rejected), and a symlinked target dir still accepting safe entries.

**6. YAML manifest fallback preserves the original error.** When `parse_frontmatter_yaml`'s markdown-link patching fallback fails too, the new code chains the original `YAMLError` as `__cause__` and wraps the retry's error in a message naming SKILL.md. Operators see what the user actually submitted instead of the post-patch garble. Test asserts `__cause__` survives.

**7. FileBrowser uses the typographic scale variable.** The inline `fontSize: "0.85rem"` in the syntax-highlighter theme is the only place in the frontend that hardcoded a rem font size — the surrounding CSS module already uses `var(--text-*)`. Replaced with `var(--text-sm)` per the CLAUDE.md font-scale rule.

### Files

| Layer | File | Notes |
|---|---|---|
| server | `api/rate_limit.py` | Proxy-aware `_client_key`, time-driven purge, `rate_limit_dep` factory |
| server | `api/registry_routes.py` | 7 closures → 7 factory calls |
| server | `api/search_routes.py` | 1 closure → 1 factory call; drop unused `Request` import |
| server | `api/auth_routes.py` | 1 closure → 1 factory call; drop unused `Request` import |
| server | `infra/cache.py` | Non-fragile `_evict_one` |
| server | `settings.py` | New `trust_proxy_hops: int = 0` field |
| server | `tests/test_api/conftest.py` | Mock settings exposes `trust_proxy_hops` |
| server | `tests/test_api/test_search_routes.py` | Same in the search-routes settings fixture |
| server | `tests/test_api/test_rate_limit.py` | +XFF tests, memory-bound tests, factory tests |
| server | `tests/test_infra/test_cache.py` | +eviction safety + soonest-expiry tests |
| shared | `dhub_core/ziputil.py` | `realpath` for symlink hardening |
| shared | `dhub_core/manifest.py` | Chain YAMLError with original cause |
| shared | `tests/test_ziputil.py` | +2 symlink tests |
| shared | `tests/test_manifest.py` | +cause-preservation test |
| frontend | `components/FileBrowser.tsx` | `var(--text-sm)` instead of `0.85rem` |

### Test plan

- [x] `pytest server/tests -m "not slow"` — 945 pass
- [x] `pytest shared/tests` — 101 pass
- [x] `pytest client/tests` — 291 pass (29 pre-existing skips)
- [x] `vitest run` (frontend) — 82/82 pass
- [x] `ruff check .` + `ruff format --check .` — clean
- [x] `mypy client/src/ server/src/ shared/src/` — clean (88 source files)
- [x] `npx tsc -b` from `frontend/` — clean
- [x] New tests cover: XFF parsing across trust levels, memory bound under 500 unique IPs, cache eviction under churn, ziputil symlink escape and safe symlinked target, manifest fallback cause chain
- [ ] CI re-runs the above on a clean checkout

### Risk / rollback

- `trust_proxy_hops` defaults to `0` so live limiting behaviour is unchanged until an operator sets it.
- The factory takes the existing `Settings` field names by string, so adding a new rate-limited endpoint is now one line instead of nine.
- TTLCache, ziputil, and manifest changes are tightening invariants, not loosening them — no caller behaviour changes for valid inputs.

https://claude.ai/code/session_01L9PFSEhRSSjec6VFeMiaGM

---
_Generated by [Claude Code](https://claude.ai/code/session_01L9PFSEhRSSjec6VFeMiaGM)_

### DIFF
--- a/frontend/src/components/FileBrowser.tsx
+++ b/frontend/src/components/FileBrowser.tsx
@@ -144,13 +144,16 @@ function TreeNodeView({
   );
 }
 
-// Custom theme based on oneDark with neon accents
+// Custom theme based on oneDark with neon accents. The syntax-highlighter
+// theme is a JS object (not a CSS module), so referencing the typographic
+// scale variable directly keeps file rendering consistent with the rest of
+// the design system per the CLAUDE.md font-scale rule.
 const neonTheme = {
   ...oneDark,
   'pre[class*="language-"]': {
     ...oneDark['pre[class*="language-"]'],
     background: "#0d0d1a",
-    fontSize: "0.85rem",
+    fontSize: "var(--text-sm)",
     lineHeight: "1.6",
   },
   'code[class*="language-"]': {
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import rate_limit_dep
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
@@ -26,16 +26,7 @@ from decision_hub.settings import Settings
 router = APIRouter(prefix="/auth", tags=["auth"])
 
 
-def _enforce_auth_rate_limit(request: Request) -> None:
-    """Rate-limit auth endpoints."""
-    state = request.app.state
-    if not hasattr(state, "_auth_rate_limiter"):
-        settings: Settings = state.settings
-        state._auth_rate_limiter = RateLimiter(
-            max_requests=settings.auth_rate_limit,
-            window_seconds=settings.auth_rate_window,
-        )
-    state._auth_rate_limiter(request)
+_enforce_auth_rate_limit = rate_limit_dep("_auth_rate_limiter", "auth_rate_limit", "auth_rate_window")
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


lfiaschi:

PR #393, opened 2026-06-09
refactor(server): rate-limit factory, fix N+1s, tighten zombie check

lfiaschi:
## Summary

Output of a deep architectural pass over the server API layer. Consolidates the most repeated boilerplate, fixes a handful of correctness/perf issues uncovered along the way, and tightens tests around the changed paths.

Net: **+338 / −166** across 10 files; **939 server + 98 shared + 291 client tests pass**, mypy + ruff clean.

### Rate limiting — one factory replaces eight copies
Every rate-limited endpoint used to define its own `_enforce_*_rate_limit` helper that lazily initialised a `RateLimiter` on `app.state` and read its limits from `settings` — 8 nearly identical copies across `registry_routes.py`, `search_routes.py` and `auth_routes.py`.

New `rate_limit_dependency(name, limit_field, window_field)` in `api/rate_limit.py` collapses the pattern into one well-documented factory. Each routes file now declares its dependencies in 1-3 lines instead of a 10-line `def`.

### `RateLimiter` no longer leaks empty entries
- Switched from `defaultdict(list)` to a plain `dict`. The old code created an empty list on every IP touch (even pre-pruning), defeating the stale-IP cleanup.
- Purge cadence now tracks **request count** (`_requests_since_purge`), not `sum(len(v) for v in store)` — the old check fired non-deterministically depending on how many timestamps any single IP had accumulated. The new test `test_stale_ips_are_purged_after_burst` proves it reliably reclaims memory after the window rolls over.

### `_check_zombie` — single DB read, not two
`GET /v1/eval-runs/{id}` used to call `find_eval_run` → `_check_zombie` → `find_eval_run` again, even when the run was healthy. Now `_check_zombie` returns the effective status and the route patches it into the already-fetched run via `_run_to_response_with_status`. Halves DB round-trips on this endpoint.

Regression guard test asserts `find_eval_run.call_count == 1` even when the run is flipped to `failed`.

### `find_eval_runs_for_version` — user filter pushed into SQL
The route used to fetch every user's runs for a version and discard foreign rows in Python. Both wasteful and a small cross-tenant exposure on the wire (we shipped UUIDs/status fields for other users' runs to a per-IP CDN cache).

`find_eval_runs_for_version(conn, version_id, *, user_id=None)` now applies the filter in the WHERE clause. The route passes `user_id=current_user.id`. Test asserts the kwarg is forwarded.

### Off-topic ask queries are now logged
When the topicality guard rejects a query (`is_skill_query=False`), `_ask_skills_inner` used to return early **without** any analytics call — so rejected queries never appeared in `search_logs` and we had no signal on what users were asking that we weren't serving. Now they're logged with an `off_topic=True` flag (no semantic change to the response).

### Sitemap cache leak fixed
`seo_routes.sitemap_xml` cached the `Response` object including its `Cache-Control: max-age=300` header. Re-serving the cached Response meant each downstream client started a fresh max-age window from *its* request time — so the browser cached for longer than intended. Now we cache the rendered XML string and build a fresh Response per request.

### `CLIVersionMiddleware` cleanup
`client_parsed` was referenced unconditionally after a conditional assignment; it only worked because `client_ver and client_parsed < ...` short-circuits on the first term. Refactored to a single early-return for the empty-header case. Pure clarity, no behaviour change.

## Test plan

- [x] `make test-server` — 939 passed, 34 deselected (slow)
- [x] `make test-shared` — 98 passed
- [x] `make test-client` — 291 passed, 29 skipped
- [x] `uvx ruff check` + `format --check` clean
- [x] `uvx mypy server/src/decision_hub/api/ shared/src/` clean
- [x] New tests:
  - `TestRateLimiter::test_reads_do_not_create_empty_entries`
  - `TestRateLimiter::test_stale_ips_are_purged_after_burst`
  - `TestRateLimitDependency` (4 tests)
  - Updated zombie + list-by-version tests assert the new call shapes

## What was deliberately left for follow-ups

These were flagged in the review but not touched here to keep the PR focused and easy to review:

1. **Granted-skill-id redundancy** — several routes call `list_granted_skill_ids` 2-3× per authenticated request because `find_skill_by_slug` / `resolve_version` each recompute it. Worth threading a request-scoped `granted_skill_ids` through, but it touches a lot of signatures.
2. **`delete_skill_version` rebuilds the S3 key** instead of returning the stored `s3_key` from `delete_version` — defensive fix, separate PR.
3. **`registry_routes.py` is 1300+ lines** even after this PR — splitting into `skills_routes`, `eval_run_routes`, and `access_routes` is a natural next step but creates a noisy diff.
4. **API tests use `MagicMock(settings)`** — adding any new setting requires updating the mock everywhere. A real `Settings(_env_file=None, **defaults)` fixture would be cleaner.

https://claude.ai/code/session_01NkEnbdcZovENNbH6idJp6s

---
_Generated by [Claude Code](https://claude.ai/code/session_01NkEnbdcZovENNbH6idJp6s)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its cross-tenant eval-run SQL filter and sitemap cache fix are catalogued in #451. Branch preserved.

PR #398, opened 2026-06-15
refactor(api): fix rate limiter GC and consolidate per-endpoint dependencies

lfiaschi:
## Summary

A scheduled deep-review pass over `server/src/decision_hub/api/` surfaced two real defects in `RateLimiter` and ~80 lines of copy/paste boilerplate spread across three route modules. This PR fixes both, with no behaviour change to legitimate traffic.

### Bug 1 — GC heuristic misfires when the map is empty

```python
# old
if total % 100 == 0:
    self._purge_stale(cutoff)
```

When the per-IP map is empty (process start, immediately after a successful purge, any window in which all entries just expired) `total == 0`, so `total % 100 == 0` is `True` and the stale-IP scan runs **on every single request**. Walking the whole map per request is the opposite of "periodic GC".

Replaced with a plain per-instance request counter so the purge only fires once every `_GC_INTERVAL` (1024) requests, regardless of how full or empty the map is.

### Bug 2 — `defaultdict` leaks ghost entries for new IPs

```python
# old
self._requests: dict[str, list[float]] = defaultdict(list)
...
timestamps = self._requests[key]   # autovivifies an empty list
self._requests[key] = [t for t in timestamps if t > cutoff]
```

Touching `defaultdict[key]` *creates* an empty list as a side effect, so every unique IP permanently grew the map by one entry until the periodic purge ran — and the periodic purge was itself broken (Bug 1). Bug 1 was masking Bug 2 by running every request, but doing it the slow way.

Switched to a plain `dict` with explicit `dict.get(key)` lookup, and to `collections.deque` so dropping expired timestamps is `O(1)` `popleft` instead of `O(n)` list rebuild.

### Bug 3 — invalid constructor arguments silently accepted

Adding `max_requests > 0` / `window_seconds > 0` validation up front. A misconfigured limiter (e.g. `max_requests=0`) previously rejected every request silently; now it fails loudly at construction.

### Cleanup — collapse 8 copies of the same boilerplate

Each rate-limited endpoint had its own `_enforce_<name>_rate_limit` function with the same 10-line shape:

```python
def _enforce_X_rate_limit(request: Request) -> None:
    state = request.app.state
    if not hasattr(state, "_X_rate_limiter"):
        state._X_rate_limiter = RateLimiter(
            max_requests=settings.X_rate_limit,
            window_seconds=settings.X_rate_window,
        )
    state._X_rate_limiter(request)
```

The new `make_rate_limit_dep(name, limit_attr, window_attr)` factory closes over those three strings and returns a FastAPI dependency that lazily caches the limiter on `app.state` exactly as before. Eight call sites collapse to one-line declarations:

```python
_enforce_publish_rate_limit = make_rate_limit_dep(
    "publish", "publish_rate_limit", "publish_rate_window"
)
```

Adding a new limited endpoint is now a single line rather than a function plus a settings entry. The lazy-init semantics, attribute name on `app.state`, and per-endpoint isolation are all preserved.

## Test plan

- [x] `make test-server` — all **945 tests pass** (`-m "not slow"`).
- [x] Extended `server/tests/test_api/test_rate_limit.py` from 5 → 17 tests covering:
  - constructor argument validation (`max_requests`, `window_seconds`),
  - the new GC interval (no purge until `_GC_INTERVAL` requests have passed),
  - the no-implicit-bucket guarantee for unknown IPs,
  - the factory's lazy initialisation, instance reuse across requests, configured-limit enforcement, and per-endpoint isolation.
- [x] `uvx ruff check` + `uvx ruff format --check` clean.
- [x] `uvx mypy` clean on the four touched source files.

https://claude.ai/code/session_014tmzi7qGGG2TXsfJonWPAC

---
_Generated by [Claude Code](https://claude.ai/code/session_014tmzi7qGGG2TXsfJonWPAC)_

### DIFF
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import make_rate_limit_dep
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
@@ -25,17 +25,7 @@ from decision_hub.settings import Settings
 
 router = APIRouter(prefix="/auth", tags=["auth"])
 
-
-def _enforce_auth_rate_limit(request: Request) -> None:
-    """Rate-limit auth endpoints."""
-    state = request.app.state
-    if not hasattr(state, "_auth_rate_limiter"):
-        settings: Settings = state.settings
-        state._auth_rate_limiter = RateLimiter(
-            max_requests=settings.auth_rate_limit,
-            window_seconds=settings.auth_rate_window,
-        )
-    state._auth_rate_limiter(request)
+_enforce_auth_rate_limit = make_rate_limit_dep("auth", "auth_rate_limit", "auth_rate_window")
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


lfiaschi:

PR #399, opened 2026-06-16
Harden zip extraction and close skill-manifest coverage gaps

lfiaschi:
## Context

I ran a deep architect-and-principal-engineer review across `client/`, `server/`, `shared/`, and `frontend/` (parallel passes for architecture/bugs, testing/observability, and security/performance). The headline finding is that the codebase is in genuinely good shape: SQL is fully parameterized, rate limiting is thread-safe, broad `except Exception` blocks consistently attach tracebacks via `logger.opt(exception=True)`, the partial index on `versions.eval_status` correctly matches the `('A','B','passed')` filter used by `find_version_for_spec`, and most "missing" FK indexes are already covered by composite primary keys or unique constraints. The big files (`database.py` at 3.5K LOC, `gauntlet.py`, `tracker_service.py`) would benefit from being split, but that's a multi-PR refactor with real regression risk and belongs in a separate effort.

So rather than ship a sprawling diff, this PR lands the three small, high-confidence fixes the review uncovered. Each is safe, scoped, and adds no new dependencies or public-API changes.

### 1. Defense-in-depth: reject symlink entries in `validate_zip_entries` (`shared/src/dhub_core/ziputil.py`)

`zipfile.extractall` itself writes a symlink entry's content as a regular file, so Python alone isn't vulnerable. But the helper is documented as the safety gate before extraction, and system `unzip` plus several Python wrapper libraries *do* honor the mode bits — they'd materialize a `link -> /etc` symlink and let later writes redirect outside the sandbox. The new check pulls the Unix file type (`S_IFMT`) from `external_attr` and rejects anything that isn't `S_IFREG` or `S_IFDIR`, but only when the entry was created on a Unix system (`create_system == 3`). Windows-created entries with bit patterns that *look* like S_IFLNK aren't misclassified, and `writestr`'s default (permission bits only, zero S_IFMT) still passes through as a regular file.

### 2. Make YAML parse failures observable (`server/src/decision_hub/domain/skill_manifest.py`)

`extract_body` and `extract_description` both had silent `except: pass` / `except: return ""` paths. When the SKILL.md frontmatter trips YAML (unquoted colons in a description are the common culprit), the regex fallback kicks in but there's no signal that the YAML path failed — so a description that silently disappears looks identical to one that was never set. Replaced both with `logger.debug` calls. Control flow is unchanged.

### 3. Close three test coverage gaps

* `extract_body` had no direct tests — it was exercised only transitively through publish.
* `extract_description`'s YAML → regex fallback path was untested. The new tests use a YAML alias (`*notyaml`) to deterministically trigger the fallback and assert the regex still finds the description.
* `repo_utils.discover_skills`, `create_zip`, and `_build_authenticated_url` had no direct unit tests (only exercised via the tracker/crawler integration suites, which spin up GitHub mocks and miss the dotfile-skip / collision-handling logic). The new tests cover the `.hidden` / `node_modules` / `__pycache__` skip behavior, the "shallowest SKILL.md wins" rule that protects checksum-based dedup, and deterministic archive ordering.

## Net change

- **+29** server-domain tests (`test_skill_manifest.py`, `test_repo_utils.py`)
- **+5** shared tests (symlink rejection, device-node rejection, regular-file-with-explicit-mode, dir-with-explicit-mode, Windows-attr-not-misclassified)
- All suites green locally: 472 server-domain tests pass (10 skipped — Gemini integration tests needing `GOOGLE_API_KEY`), 103 shared tests pass, ruff clean, mypy clean
- No public-API changes, no new dependencies, no migrations

## Review findings deliberately left for follow-ups

These came up but didn't make the cut here because they need design discussion, carry regression risk, or were on inspection not actually problems:

- **Split `database.py` (3.5K LOC)** into `tables.py` / `queries.py` / `inserts.py` / `updates.py`. Real cohesion problem, but a multi-PR refactor.
- **Rate limiter per-IP only** — claimed as bypassable via botnet. True in theory, but the auth bottleneck (10 req/60s) is intentionally cheap; moving to per-user-id requires settling identity-before-auth semantics and is a design call, not a fix.
- **JWT carries `github_orgs` snapshot** — a removed org member keeps access for the token's lifetime. Real, but the right fix (live membership check on every request) needs perf evaluation against the current snapshot model.
- **"Missing" indexes** flagged by the review (`user_api_keys.user_id`, `eval_runs.user_id`, `skills.org_id`) — on inspection, all are already covered by composite unique constraints or existing composite indexes.
- **Per-container `TTLCache` not invalidated on publish** — by design: TTLs are 60s and the cache is per-Modal-replica, so invalidating on the writing container wouldn't help the others anyway.

Test plan:
- [x] `pytest shared/tests/` — 103 passed
- [x] `pytest server/tests/test_domain/` — 472 passed, 10 skipped
- [x] `ruff check .` — All checks passed
- [x] `ruff format --check .` — 198 files formatted
- [x] `mypy shared/src server/src/decision_hub/domain/skill_manifest.py` — no issues

https://claude.ai/code/session_01Fm2H4MYLBTFpcKqdvw5KcR

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fm2H4MYLBTFpcKqdvw5KcR)_

### DIFF
--- a/server/src/decision_hub/domain/skill_manifest.py
+++ b/server/src/decision_hub/domain/skill_manifest.py
@@ -8,6 +8,7 @@ functions: extract_body, extract_description, parse_eval_cases_from_zip.
 import re
 
 import yaml
+from loguru import logger
 
 from decision_hub.models import EvalCase
 from dhub_core.manifest import (  # noqa: F401
@@ -26,7 +27,8 @@ def extract_body(content: str) -> str:
     try:
         _, body = split_frontmatter(content)
         return body
-    except ValueError:
+    except ValueError as exc:
+        logger.debug("extract_body: frontmatter split failed ({}), returning empty body", exc)
         return ""
 
 
@@ -40,7 +42,8 @@ def extract_description(content: str) -> str:
     """
     try:
         frontmatter_str, _ = split_frontmatter(content)
-    except ValueError:
+    except ValueError as exc:
+        logger.debug("extract_description: frontmatter split failed ({})", exc)
         return ""
 
     # Try standard YAML parsing first
@@ -49,8 +52,11 @@ def extract_description(content: str) -> str:
         if isinstance(data, dict):
             desc = data.get("description")
             return str(desc) if desc else ""
-    except yaml.YAMLError:
-        pass
+    except yaml.YAMLError as exc:
+        # Frontmatter contains values YAML can't parse (e.g. unquoted
+        # colons in the description).  Fall through to the regex path
+        # rather than dropping the field on the floor.
+        logger.debug("extract_description: YAML parse failed ({}), falling back to regex", exc)
 
     # Fallback: extract description line directly via regex.
     # Handles values with unquoted YAML-special characters like colons.
--- a/server/tests/test_domain/test_repo_utils.py
+++ b/server/tests/test_domain/test_repo_utils.py
@@ -1,0 +1,143 @@
+"""Unit tests for decision_hub.domain.repo_utils — discover_skills, create_zip,
+and _build_authenticated_url. clone_repo is exercised in integration suites
+because it shells out to git."""
+
+import io
+import zipfile
+from pathlib import Path
+
+import pytest
+
+from decision_hub.domain.repo_utils import (
+    _build_authenticated_url,
+    create_zip,
+    discover_skills,
+)
+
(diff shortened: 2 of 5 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Zip hardening covered by retained #438. Branch preserved.

PR #400, opened 2026-06-17
Targeted refactor: JWT 500→401, CLI subprocess timeouts, frontend fetch timeout

lfiaschi:
## Context

Output of a scheduled deep-review routine. Three exploration agents (server, client, frontend) audited the codebase end-to-end; the high-level findings are summarized below and the most actionable, low-risk items are fixed here. The rest of the assessment is captured for follow-up work but not implemented in this PR — the goal was tight, testable fixes, not a sprawling sweep.

## What's in this PR

Four small commits, one theme each:

### 1. `server(deps): treat malformed JWT claims as 401 instead of 500`
A signed JWT missing `sub` / `username`, or with a non-UUID `sub`, used to raise `KeyError` / `ValueError` from `UUID(payload["sub"])` and escape into FastAPI's default 500 handler — leaking a stack trace on debug builds for attacker-controlled input.

- `server/src/decision_hub/api/deps.py`: both `get_current_user` (returns 401) and `get_current_user_optional` (returns None) now guard the final `User(...)` construction.
- New tests in `server/tests/test_api/test_deps.py` cover the three malformed-payload shapes.

### 2. `client(cli): cap subprocess probes; narrow swallowed exceptions`
Three silent-failure modes in the CLI rolled into one commit:

- `_detect_installer` / `_query_version` / `_upgrade` in `client/src/dhub/cli/app.py` previously had **no timeout** on the uv / pipx / pip subprocess calls. A hung package manager would freeze the terminal before the user even saw a `--version` prompt. They now go through a shared `_run_probe` helper (10s probe cap, 5min upgrade cap, clean timeout/missing-binary handling).
- `_detect_branch` in `client/src/dhub/cli/registry.py` caught `except Exception:`. Narrowed to the four expected error types plus a 5s git timeout, so coding bugs surface instead of returning `"main"` silently.
- `_read_cache` / `_write_cache` in `client/src/dhub/cli/version_check.py` swallowed every exception. Now narrowed to `OSError` / `JSONDecodeError` / `UnicodeDecodeError` with explicit type guards on the parsed payload. The outer `show_update_notice` guard stays broad — that's the right contract for a non-essential best-effort feature.
- New `client/tests/test_cli/test_app.py` and additions to `test_version_check.py` cover the timeout / missing-binary / malformed-cache paths.

### 3. `frontend(api): time out JSON requests; stabilize chat message keys`
`fetchJSON` in `frontend/src/api/client.ts` had no timeout — a hung backend left the UI spinning forever. Wraps every call in an `AbortController` with a 45s cap (high enough to cover the slowest endpoint, `/v1/ask`, which uses an Anthropic judge). Caller-supplied `signal`s are still honored and either signal aborting wins.

Also gives `AskModal` message list a self-describing `${role}-${index}` key — append-only today, but the named key makes intent explicit if we later add edit/delete.

### 4. `chore: refresh uv.lock to match pyproject versions`
Side-effect of running `uv run`: `dhub-cli 0.12.1→0.12.3` and `dhub-core 0.2.8→0.2.10` had been bumped in pyproject.toml without a lockfile update.

## Test plan

- [x] `uv run --package dhub-cli pytest client/tests/` — **304 passed, 29 skipped**
- [x] `DHUB_ENV=dev uv run --package decision-hub-server pytest server/tests/ -m "not slow"` — **936 passed, 34 deselected**
- [x] `cd frontend && npx vitest run` — **82 passed, 8 files**
- [x] `cd frontend && npx tsc -b` — clean
- [x] `cd frontend && npx eslint src/api/client.ts src/components/AskModal.tsx` — clean

## Review findings deferred to follow-ups

The agents flagged additional items that warrant separate, larger PRs (intentionally not in scope here):

- **Rate limiter is per-container on Modal.** A distributed attacker can multiply the per-IP limit by the replica count. Needs either Redis-backed limits or an explicit documentation+settings change.
- **`backfill_embeddings.py` / `crawler/__main__.py` have no checkpointing.** A crash mid-run reprocesses everything. CLAUDE.md already mandates incremental + resumable + parallel for batch scripts; these don't comply.
- **`gemini.py` retry uses `time.sleep`.** Fine on Modal's sync FastAPI workers, but worth migrating to `httpx.AsyncClient` if/when the server moves to async handlers.
- **`storage.py` / `anthropic_client.py` timeouts.** Anthropic already has `timeout=60`; S3 calls inherit boto3 defaults (no per-op cap). A targeted PR could pin both end-to-end.
- **`SkillDetailPage.tsx` JSZip retry / in-flight cancellation.** The retry currently relies on a `useEffect` re-trigger via state flag, which works but is brittle to refactor; an explicit `AbortController` per skill load would be cleaner.
- **HomePage / SkillsPage missing `@media (max-width: 480px)` rules.** CLAUDE.md mandates mobile-first; a few pages still ship desktop-only.

https://claude.ai/code/session_01CB6xavzmcDUZhNrKDr81QG

---
_Generated by [Claude Code](https://claude.ai/code/session_01CB6xavzmcDUZhNrKDr81QG)_


lfiaschi:

PR #402, opened 2026-06-19
fix: deterministic pagination + batched orphaned tracker updates

lfiaschi:
## Why

I did a deep architectural pass over the server hot paths (`database.py`, `tracker_service.py`, `publish_pipeline.py`, `search_routes.py`). The pass surfaced four issues that all violate rules the project sets for itself in `CLAUDE.md`:

- *"Every query with `LIMIT` must have an explicit `ORDER BY` with a unique tiebreaker"*
- *"Don't advance state on failure"*
- *"Isolate side operations"*

This PR fixes them in one focused commit, with tests. I picked the changes ruthlessly — there are other findings I deliberately left for follow-up PRs (shared `httpx.AsyncClient` lifecycle, magic-number truncation constants, mobile breakpoint audit) because they each warrant their own review.

### 1. Deterministic pagination (correctness, P1)

Five queries had `ORDER BY <non-unique-column>` followed by `LIMIT`, which makes result ordering — and therefore pagination cuts — **undefined** when ties exist. Postgres is free to return tied rows in any order, and in practice it changes between requests as soon as the planner picks a different path.

| Function | File:line | Old | New |
|---|---|---|---|
| `search_skills_hybrid` (FTS) | `database.py:2043` | `fts_rank DESC` | `fts_rank DESC, org_slug, skill_name` |
| `search_skills_hybrid` (vector) | `database.py:2055` | `vec_dist ASC` | `vec_dist ASC, org_slug, skill_name` |
| `fetch_similar_skills` | `database.py:2128` | `vec_dist ASC` | `vec_dist ASC, org_slug, skill_name` |
| `find_active_eval_runs_for_user` | `database.py:2727` | `created_at DESC` | `created_at DESC, id DESC` |
| `list_tracker_metrics` | `database.py:3252` | `recorded_at DESC` | `recorded_at DESC, id DESC` |

Skills can plausibly share an FTS rank (short manifests, similar wording); metrics rows written in the same batch can share `recorded_at` down to the microsecond. Either caused the same query to return rows in a different order across requests, silently shifting page boundaries.

### 2. Visibility on partial tracker failure (correctness, P1)

`process_tracker` cleared `last_error` whenever *any* skill in a multi-skill repo published successfully, even when others failed in the same commit. Operators only saw the errors in transient logs — never in the DB or UI — so broken skills sat stale until the next commit happened to touch them.

The intentional design ("don't replay the same commit") still holds: SHA still advances on partial success. But errors are now persisted on every failure, partial or total, so operators see the broken skills.

```diff
- last_error="; ".join(errors)[:500] if all_failed else None,
+ error_msg = "; ".join(errors)[:500] if errors else None
+ ...
+ last_error=error_msg,
```

Two existing tests pinned the old silent-clear behavior; updated them to assert the new (correct) behavior with comments explaining why.

### 3. Batched orphaned-tracker error writes (perf, P1)

`_persist_orphaned_tracker_errors` looped over orphaned IDs calling `update_skill_tracker` per ID — N round-trips when a whole Modal batch dies (which is exactly when this code runs). Added `batch_mark_trackers_orphaned` in `database.py` (mirrors the existing `batch_set_tracker_errors`, `batch_defer_trackers`, etc. pattern) and switched the call site to one batched UPDATE.

```diff
- with engine.connect() as conn:
-     for tid in orphaned_ids:
-         update_skill_tracker(conn, tid, last_checked_at=now, last_error=error_msg)
-     conn.commit()
+ with engine.connect() as conn:
+     batch_mark_trackers_orphaned(conn, list(orphaned_ids), checked_at=now, error_message=error_msg)
+     conn.commit()
```

Updated the existing test to enforce that exactly one execute happens regardless of orphan count.

### 4. Better log context on scan-report side-failures (DX, P2)

`_store_scan_result_safely` in `publish_pipeline.py` still swallows failures (intentional per CLAUDE.md's "isolate side operations" rule — publish must not block on Cisco scan storage), but the warning now includes `semver` and `version_id` so an orphaned scan can be re-attached via `backfill_scan_reports.py` without grepping S3.

## Verification

- `uvx ruff check` + `format --check` clean on the touched files
- `uvx mypy` clean on the touched files
- `pytest server/tests/ -m "not slow"` — **936 passed, 0 failures, 0 regressions**
- New / updated tests:
  - `TestBatchMarkTrackersOrphaned` — asserts a single `execute` regardless of N
  - `TestListTrackerMetrics::test_orders_by_recorded_at_with_id_tiebreaker` — asserts the compiled SQL contains the tiebreaker
  - `test_persist_orphaned_tracker_errors` — rewritten for batched API
  - `test_partial_success_advances_sha` / `test_three_of_five_succeed_advances_sha_clears_error` — pinned to new visibility-preserving behavior

## Test plan

- [ ] CI green on `lint`, `typecheck`, `test-server`, `migrate-check`, `schema-drift`
- [ ] Spot-check on dev that `/v1/skills/ask` still returns stable result ordering for a query that produces ties (e.g. very generic terms — try `q=skill`)
- [ ] Spot-check `/v1/trackers/metrics` returns identical ordering across two back-to-back requests

## Out of scope (follow-up candidates)

These showed up in the audit but warrant separate PRs:

- **Shared `httpx.AsyncClient` lifecycle** — `search_routes.py:257,299` creates a fresh client per request, defeating connection pooling. Needs lifespan wiring + a dependency.
- **Truncation magic numbers** — `[:500]`, `[:200]`, `[:100]` scattered across `gauntlet.py`, `tracker_service.py`, `evals.py`, `crawler/processing.py`. Worth a `MAX_ERROR_MSG_LEN` constant.
- **Mobile-first audit** — `OrgDetailPage.tsx` has inline `fontSize: "4rem"` 404 heading that's unreadable at 400px. `SkillsPage.module.css` mobile breakpoint starts at 640px instead of 480px per the CLAUDE.md rule.
- **Authorization matrix tests** — no parameterized tests for `(anon, org-member, non-member) × (public, org, private)` visibility paths.

---

Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_018stKPHcR6ZBtE2FrX6bgiE

---
_Generated by [Claude Code](https://claude.ai/code/session_018stKPHcR6ZBtE2FrX6bgiE)_

### DIFF
--- a/server/src/decision_hub/domain/publish_pipeline.py
+++ b/server/src/decision_hub/domain/publish_pipeline.py
@@ -841,8 +841,13 @@ def _try_store_scan_result(
             )
             fresh_conn.commit()
     except Exception:
+        # Side-operation: never block publish on scan-storage failure (CLAUDE.md
+        # "Isolate side operations" rule). Capture enough identifiers so the
+        # orphaned scan can be backfilled from logs via backfill_scan_reports.
         logger.opt(exception=True).warning(
-            "Failed to store scan report for {}/{} — scan data lost but publish succeeded",
+            "Failed to store scan report for {}/{} version={} version_id={} — scan data lost but publish succeeded",
             org_slug,
             skill_name,
+            version,
+            version_id,
         )
--- a/server/src/decision_hub/domain/tracker_service.py
+++ b/server/src/decision_hub/domain/tracker_service.py
@@ -613,30 +613,34 @@ def _persist_orphaned_tracker_errors(
     *,
     error_msg: str,
 ) -> None:
-    """Update last_error for trackers whose Modal containers died before writing a result."""
-    from decision_hub.infra.database import update_skill_tracker
+    """Update last_error for trackers whose Modal containers died before writing a result.
+
+    Issues a single batched UPDATE rather than one per tracker so a wave of
+    Modal failures doesn't translate into N database round-trips.
+    """
+    from decision_hub.infra.database import batch_mark_trackers_orphaned
 
     if not orphaned_ids:
         return
 
     now = datetime.now(UTC)
     tracker_by_id = {t.id: t for t, _ in changed_trackers}
+    for tid in orphaned_ids:
+        tracker = tracker_by_id.get(tid)
+        logger.warning(
+            "tracker_id={} repo={} status=orphaned_failure error={}",
+            tid,
+            tracker.repo_url if tracker else "?",
+            error_msg,
+        )
     try:
         with engine.connect() as conn:
-            for tid in orphaned_ids:
-                tracker = tracker_by_id.get(tid)
-                logger.warning(
-                    "tracker_id={} repo={} status=orphaned_failure error={}",
-                    tid,
-                    tracker.repo_url if tracker else "?",
-                    error_msg,
-                )
-                update_skill_tracker(
-                    conn,
-                    tid,
-                    last_checked_at=now,
-                    last_error=error_msg,
-                )
+            batch_mark_trackers_orphaned(
+                conn,
+                list(orphaned_ids),
+                checked_at=now,
+                error_message=error_msg,
+            )
             conn.commit()
     except Exception:
         logger.opt(exception=True).error("Failed to persist orphaned tracker errors")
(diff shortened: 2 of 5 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its batch_mark_trackers_orphaned helper is catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #403, opened 2026-06-20
refactor(api): consolidate rate-limit deps and fix two zombie-sweep bugs

lfiaschi:
## Why

I started this branch as an architecture review of the server. Three findings were small, mechanical, and tightly related, so they're bundled here. Larger findings (splitting `database.py` and `cli/registry.py`, moving rate-limit state to Postgres/Redis for cross-container fairness, etc.) are out of scope and worth their own issues.

### 1. Rate-limit factory — deletes ~80 LOC of duplication

`auth_routes.py`, `search_routes.py`, and `registry_routes.py` each had near-identical `_enforce_*_rate_limit` blocks — 8 copies total, all doing the same lazy-init dance on `app.state`. Replaced them with a single `make_rate_limit_dep(name)` factory in `rate_limit.py` that reads `settings.{name}_rate_limit` / `{name}_rate_window` and caches the limiter on `_{name}_rate_limiter`. Adding a new rate-limited endpoint is now a one-liner.

```python
```

### 2. `RateLimiter` perf bug — `O(N_ips)` on every request

The old purge heuristic recomputed `sum(len(v) for v in self._requests.values())` on every `__call__` to decide when to sweep stale IPs. With even modest traffic that walked every bucket on every rate-limited request. The `total % 100` trigger was also brittle: right after a sweep, `total == 0` (which is `0 % 100 == 0`), so the next request swept again.

Replaced with an `O(1)` call counter plus a configurable `purge_interval` (default 256). The previous test suite did not exercise the sweep at all — added a test for it.

### 3. Eval-run zombie sweep — two bugs at once

`registry_routes._check_zombie` returned the new status as a string, then forced the route handler to re-read the row to surface the updated `error_message` and `completed_at`. That's an extra DB roundtrip per `/v1/eval-runs/{id}` and `/v1/eval-runs/{id}/logs` call — both polled while an eval is running.

```python
# before
_check_zombie(conn, run)
run = find_eval_run(conn, parsed_id)  # always runs, even if nothing changed
return _run_to_response(run)
```

`_check_zombie` now returns the updated `EvalRun` via `dataclasses.replace` so callers don't re-query.

While there, fixed a related correctness bug: `update_eval_run_status` unconditionally set `heartbeat_at = now()`, so the API write that *records a worker's death* also refreshed its liveness signal — masking the next zombie-detection cycle. Added `bump_heartbeat: bool = True` (workers keep the default, the zombie sweep passes `False`). As a side effect the function now skips the UPDATE entirely when there's nothing to write.

## Tests

- 6 new tests for the factory + purge behaviour (`test_rate_limit.py`).
- New `test_infra/test_eval_runs_db.py` covers the three branches of `update_eval_run_status`: default bump, no bump, no-op.
- Existing zombie test updated to assert the new single-read contract and the `bump_heartbeat=False` invariant.

## Verification

- Full server suite: **941 passed, 0 failed** (`pytest -m "not slow"`).
- `ruff check`, `ruff format --check`, and `mypy` clean on all touched files.
- No DB migrations, no behavioural changes to public APIs.

## Follow-ups (intentionally not in this PR)

- `infra/database.py` is 3,465 lines (table defs + 50+ query functions + row mappers); should split per aggregate.
- `cli/registry.py` is 1,912 lines; same treatment by command group.
- In-memory rate limiter is per Modal container — single client can burn each container's allowance independently. Worth moving to Postgres/Redis if abuse becomes a real concern.
- `get_eval_run_logs` reads S3 chunks sequentially; trivial to parallelise with `ThreadPoolExecutor`.

https://claude.ai/code/session_015oiaKCbzhPtXEe5g8sw1sk

---
_Generated by [Claude Code](https://claude.ai/code/session_015oiaKCbzhPtXEe5g8sw1sk)_

### DIFF
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import make_rate_limit_dep
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
@@ -26,16 +26,7 @@ from decision_hub.settings import Settings
 router = APIRouter(prefix="/auth", tags=["auth"])
 
 
-def _enforce_auth_rate_limit(request: Request) -> None:
-    """Rate-limit auth endpoints."""
-    state = request.app.state
-    if not hasattr(state, "_auth_rate_limiter"):
-        settings: Settings = state.settings
-        state._auth_rate_limiter = RateLimiter(
-            max_requests=settings.auth_rate_limit,
-            window_seconds=settings.auth_rate_window,
-        )
-    state._auth_rate_limiter(request)
+_enforce_auth_rate_limit = make_rate_limit_dep("auth")
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


lfiaschi:

PR #406, opened 2026-06-23
refactor: harden rate limiter, DRY 9 wrappers, fix CLI token mode + SQL tiebreakers

lfiaschi:
## Summary

A principal-engineer review pass: seven focused fixes from a broader audit, all small, individually reviewable, and covered by tests. The bigger structural carve-outs surfaced by the review (`infra/database.py` at 3,465 lines, `api/registry_routes.py` at 1,352 lines, the in-memory rate-limiter/cache that silently become per-replica state once you scale Modal) are tracked for follow-up PRs — they'd dwarf this diff and need their own discussion.

Net: **+291 / -125** lines across **14 files**.

### Bug — Rate limiter collapsed all traffic behind the Modal edge to one bucket

`api/rate_limit.py` and `api/deps.py` keyed on `request.client.host`. Under Modal (and any other proxy/CDN that terminates TCP) that's the proxy IP — identical for every caller. One noisy client could lock out every other tenant sharing the edge.

New `client_ip(request)` helper honours `X-Forwarded-For` (leftmost = original client), falls back to `X-Real-IP`, then the direct peer, then `"unknown"`. Both the `RateLimiter` and the JWT logger now use it. Regression test covers the exact scenario: two clients with different XFF behind the same peer IP no longer share a bucket.

### Bug — Search ORDER BYs lacked tiebreakers

`search_skills_hybrid` (FTS + vector) and `fetch_similar_skills` ordered by `fts_rank`/`vec_dist` alone — equal scores produced non-deterministic paging, which the project SQL rule explicitly bans. Added `(org_slug, skill_name)` tiebreakers on all three `LIMIT`'d queries.

### Hardening — CLI bearer token was world-readable on permissive umasks

`Path.write_text` uses the process umask, so on systems with a permissive default (CI containers, shared hosts) `~/.dhub/config.{env}.json` ended up `0644` — the long-lived bearer token was readable by every local user. `save_config` now chmods the dir to `0700` and the file to `0600`. Best-effort (suppresses `OSError`) so Windows and read-only mounts still save successfully.

### Hardening — `/v1/stats` was the only public read with no rate limit

Now shares the existing `list_skills` per-IP limiter — both are lightweight cached reads from the same DB.

### Code health — DRY the 9 `_enforce_*_rate_limit` wrappers

Nine near-identical 9-line copies across `registry_routes`, `search_routes`, and `auth_routes` collapsed into one factory:

```python
_enforce_publish_rate_limit = rate_limiter_dep("publish")  # reads settings.publish_rate_limit/window
```

Adding a new limited endpoint is now one line of wiring plus the two `Settings` fields. Also dropped a stacked `# noqa: F401  # noqa: F401` on the registry_service re-exports.

### Frontend / API contract — `OrgStatsResponse.total`

`OrgsPage` used `orgs.length` as the "N organizations found" header, violating the project React rule. The type didn't even expose `total`. Server now returns `total: int` (currently `len(items)`; field exists so adding server-side pagination doesn't require a frontend type change), and the UI reads it.

## What's NOT in this PR (follow-ups identified by the review)

1. **Split `infra/database.py`** (3,465 lines) into `database/{skills,versions,trackers,evals,search,orgs}.py`. Mechanical but every import path moves — staged over 2–3 PRs with tests written first.
2. **Split `api/registry_routes.py`** (1,352 lines) by resource (skills / versions / eval-runs / access-grants / audit).
3. **`api/registry_service.py` re-export pile**. Removing the re-exports breaks 60+ test patches plus `modal_app.py`, `crawler/processing.py`, and 3 backfill scripts. Worth a dedicated PR.
4. **In-memory rate limiter + TTL cache become per-Modal-container state once `max_inputs>1` triggers a second replica.** Document the limit, add Redis-backed alternates before the fleet grows.
5. **CLI `registry.py` is 1,912 lines** mixing HTTP, zip, git, version logic across 10 commands. Worth splitting + introducing a shared `httpx.Client` (today every call opens a fresh client — 50+ sites).
6. **JWT `github_orgs` claim is stale for up to 1 year** after a membership change. Shorter TTL or a `version` claim.

## Test plan

- [x] `uvx mypy client/src/ server/src/ shared/src/` — 88 files, no issues
- [x] `pytest server/tests/ -m "not slow"` — **944 passed** (+16 new rate-limit tests)
- [x] `pytest client/tests/` — 292 passed, 29 skipped (incl. new chmod test, skipped on Windows)
- [x] `pytest shared/tests/` — 98 passed
- [x] `cd frontend && npx tsc -b` — clean
- [x] `cd frontend && npx vitest run` — 82/82
- [x] `cd frontend && npx eslint .` — 1 pre-existing warning, unrelated
- [ ] Smoke-test on dev once approved: capture a real `/v1/skills` request, confirm the new XFF logging shows the original client IP and not the Modal proxy IP

https://claude.ai/code/session_01HqU6JuJT2YFjH8tezZeBJP

---
_Generated by [Claude Code](https://claude.ai/code/session_01HqU6JuJT2YFjH8tezZeBJP)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -80,16 +81,27 @@ def load_config() -> CliConfig:
 
 
 def save_config(config: CliConfig) -> None:
-    """Save CLI config to ~/.dhub/config.{env}.json.
+    """Save CLI config to ``~/.dhub/config.{env}.json``.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ``~/.dhub`` directory if needed and tightens permissions
+    on both the directory and the file to user-only (``0700``/``0600``).
+    The config holds a long-lived bearer token, so on multi-user systems
+    or CI containers with permissive umasks the default ``0644`` would
+    leave the token world-readable. ``chmod`` is a no-op on Windows.
     """
     CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+    # ``chmod`` is best-effort: on Windows or a read-only mount it can
+    # legitimately fail without affecting save correctness, so we
+    # swallow OSError rather than aborting the save.
+    with contextlib.suppress(OSError):
+        CONFIG_DIR.chmod(0o700)
     path = config_file()
     path.write_text(
         json.dumps(asdict(config), indent=2) + "\n",
         encoding="utf-8",
     )
+    with contextlib.suppress(OSError):
+        path.chmod(0o600)
 
 
 def get_api_url() -> str:
--- a/client/tests/test_cli/test_config.py
+++ b/client/tests/test_cli/test_config.py
@@ -91,6 +91,30 @@ class TestSaveConfig:
         assert loaded.orgs == ("alice", "pymc-labs")
         assert loaded.default_org == "pymc-labs"
 
+    def test_file_and_dir_permissions_are_tightened(self, tmp_path, monkeypatch):
+        """Saved config and parent dir must be user-only (0600/0700).
+
+        The config file holds a long-lived bearer token. On systems with
+        a permissive umask (CI containers, shared hosts) the default file
+        mode would be world-readable, leaking the token.
+        """
+        import platform
+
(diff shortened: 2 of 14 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its OrgStatsResponse.total fix is catalogued in #451. Branch preserved.

PR #407, opened 2026-06-24
Audit fixes: tracker mapper, zip-bomb, rate-limit dedup, public-surface tightening

lfiaschi:
## Why

Synthesis of a multi-dimension codebase audit (architecture, bugs, security, performance, testing). Each change is small, justified by a concrete failing scenario or measurable cost, and verified by existing tests plus new regressions. Tier-1 quick wins only — bigger structural work is called out below for follow-up.

### Bug fixes

| # | Where | What |
|---|---|---|
| 1 | `infra/database.py` | `_row_to_skill_tracker` was a hand-rolled mapper that **silently dropped `consecutive_permanent_failures`** — every consumer read 0 regardless of DB state, breaking the disable-after-N-failures policy. Now derives fields from `SkillTracker.__dataclass_fields__` via `row._mapping`, so future columns are picked up automatically. Regression test added. |
| 2 | `domain/tracker_service.py` | Multi-skill tracker repos cleared `last_error` whenever *any* skill succeeded. Operators couldn't see which sub-skills regressed. Now records the partial-failure summary whenever `errors` is non-empty. Two tracker_service tests updated. |
| 3 | `scripts/activate_trackers.py` | Loop ran inside a single `engine.connect()`. An `IntegrityError` mid-loop's `conn.rollback()` discarded every prior uncommitted insert; the final `commit()` overstated the `created` count. Now commits per row. |
| 4 | `domain/publish_pipeline.py` | If `run_eval.spawn(...)` raised, the `eval_run` row sat in `pending` forever because the zombie detector only covers `running/judging/provisioning`. CLI tail loops hung. Now marks the row `failed` with the spawn error before re-raising. |
| 5 | `infra/database.py` | `find_audit_logs(limit=None, offset=N)` issued an unbounded fetch. Combined with the 30 s `statement_timeout` it could still OOM the container. Capped at 1000. |
| 6 | `api/registry_routes.py` | `get_eval_run` called `find_eval_run` twice per poll (the UI polls). `_check_zombie` now returns the new `(status, error_message, completed_at)` so the caller patches the dataclass in-process — halves DB round-trips on the hot path. |

### Security

| # | Where | What |
|---|---|---|
| 7 | `domain/publish.py` | **Zip-bomb fix.** The old check summed `info.file_size` from the central directory, which is attacker-controlled metadata. A lying header could let an over-budget payload reach `zf.read(name)` unbounded. The fix reads each member via `zf.open(name).read(_MAX_FILE_SIZE + 1)` with a running remaining-budget counter; per-file *and* total caps are enforced against actual decompressed bytes. |
| 8 | `api/registry_routes.py` + frontend | Public `AuditLogResponse` and the SkillDetail page were exposing `quarantine_s3_key` — an internal S3 path with no value to external callers and an info-disclosure on rejected publish attempts. Removed from server response model, frontend type, frontend display, and orphaned CSS. Test asserts the field is absent from the response. |
| 9 | `api/keys_routes.py` | `StoreKeyRequest` had no length caps. An authenticated user could POST a 10 MB blob to be Fernet-encrypted, persisted, and re-decrypted in eval-worker memory. Added `Field(max_length=128)` / `Field(max_length=8192)`. Two new 422-response tests. |
| 10 | `api/registry_routes.py` | Added rate limiters to previously-unprotected public endpoints — `/v1/stats`, `summary`, `latest-version`, both `eval-report` routes — via the new factory. |

### Performance

| # | Where | What |
|---|---|---|
| 11 | `infra/storage.py` | boto3 was created without a `Config`. Defaults: `connect_timeout=60s`, `read_timeout=60s`, **5 legacy retries** — a single stalled S3 call could wedge a worker for ~5 minutes, on a `max_inputs=100` Modal replica that's a likely DoS amplifier. Now configured with `connect=3`, `read=15`, `retries=standard/3`. `modal_app.py` inherits via `create_s3_client`. |

### Simplification (no behaviour change)

| # | Where | What |
|---|---|---|
| 12 | `api/rate_limit.py` | New `make_rate_limit_dependency(name)` factory replaces **nine** near-identical `_enforce_*_rate_limit` lazy-init dependencies across `registry_routes`, `search_routes`, `auth_routes`. ~60 LOC of boilerplate removed. The settings naming convention `{name}_rate_limit` / `{name}_rate_window` is now load-bearing. |
| 13 | `domain/publish_pipeline.py` | Collapsed five `_build_*_fn` Gemini closures (~100 LOC) into a single `_build_gemini_fn(settings, gemini, fn_name)` factory via `getattr`. Kept the five named one-line wrappers because ~30 patch decorators in route tests target them by name. |
| 14 | `api/registry_service.py` | Pruned dead `# noqa: F401` re-exports — eleven symbols that nothing in `server/`, `client/`, `scripts/`, or tests actually imported from this module (tests patch them at their domain location). Kept `classify_skill_category` and `run_gauntlet_pipeline` (real callers in scripts/backfills/crawler). |

### Testing

| # | Where | What |
|---|---|---|
| 15 | `tests/test_infra/test_cache.py` | Replaced five `time.sleep(0.01–0.02)` TTL-expiry tests with a `fake_clock` fixture that patches the cache module's `time` reference. Removes flakiness on busy CI runners and ~100 ms of wall-clock from the suite. |
| 16 | `tests/test_infra/test_tracker_db.py` | Tracker-row mock now carries a dict-shaped `_mapping` so the new `_row_to_skill_tracker` can iterate it; added regression test for `consecutive_permanent_failures`. |

## Verification

- **Server tests:** 931 passed, 24 deselected (`-m "not slow"`, excluding the 3.6k-line `test_arxiv_gauntlet` benchmark).
- **Client tests:** 291 passed, 29 skipped.
- **Shared tests:** 98 passed.
- **Frontend tests:** 82 passed.
- **Lint:** ruff check + format clean.
- **Frontend typecheck:** clean.

## Deferred to follow-up PRs (called out from the audit, not in scope here)

- Split `infra/database.py` (3,465 LOC) and `api/registry_routes.py` (1,352) along the section banners they already use.
- Replace the remaining 13 hand-rolled `_row_to_*` mappers with the same `_mapping`-based pattern (#1 establishes the template).
- Audit-log HMAC chain — today's `checksum` column is dedup-only despite the naming.
- Reclassify `test_arxiv_gauntlet.py` as a benchmark; introduce a `make bench` target.
- OpenTelemetry traces and a `/metrics` endpoint — currently zero metrics/tracing instrumentation.
- Rate-limiter keyed by `X-Forwarded-For` — Modal terminates TLS at the edge, so today every per-IP limiter sees the proxy IP.
- JWT default expiry: drop from 1 year to 24 h + revocation table (breaking; needs CLI coordination).
- Rename eval-log S3 chunks to include max `event_seq` so `/eval-runs/{id}/logs` polling becomes O(1) per delta instead of O(N²).

https://claude.ai/code/session_01MyGu22TqzGKBAVURr3Qr88

---
_Generated by [Claude Code](https://claude.ai/code/session_01MyGu22TqzGKBAVURr3Qr88)_

### DIFF
--- a/frontend/src/pages/SkillDetailPage.module.css
+++ b/frontend/src/pages/SkillDetailPage.module.css
@@ -613,16 +613,6 @@ a.metaItem:hover {
   background: rgba(255, 230, 0, 0.08);
 }
 
-.auditQuarantine {
-  font-family: 'Share Tech Mono', monospace;
-  font-size: var(--text-xs);
-  color: var(--neon-pink);
-  padding: 4px 8px;
-  background: rgba(255, 45, 149, 0.08);
-  border-radius: 4px;
-  align-self: flex-start;
-}
-
 /* Similar Skills panel */
 .similarHeader {
   font-family: 'Rajdhani', sans-serif;
--- a/frontend/src/pages/SkillDetailPage.tsx
+++ b/frontend/src/pages/SkillDetailPage.tsx
@@ -771,12 +771,6 @@ function AuditTab({
             {entry.check_results.length > 0 && (
               <CheckResultsGrid checks={entry.check_results} />
             )}
-
-            {entry.quarantine_s3_key && (
-              <span className={styles.auditQuarantine}>
-                Quarantined: {entry.quarantine_s3_key}
-              </span>
-            )}
           </div>
         </NeonCard>
       ))}
--- a/frontend/src/test/helpers.tsx
+++ b/frontend/src/test/helpers.tsx
@@ -107,7 +107,6 @@ export function makeAuditLogEntry(
     ],
     llm_reasoning: null,
     publisher: "dev@example.com",
-    quarantine_s3_key: null,
     created_at: "2025-01-01T00:00:00Z",
     ...overrides,
   };
--- a/frontend/src/types/api.ts
+++ b/frontend/src/types/api.ts
@@ -92,7 +92,6 @@ export interface AuditLogEntry {
   check_results: CheckResult[];
   llm_reasoning: Record<string, unknown> | null;
   publisher: string;
-  quarantine_s3_key: string | null;
   created_at: string | null;
 }
 
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import make_rate_limit_dependency
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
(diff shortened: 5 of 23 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its zip-bomb actual-bytes fix (the only correct variant) and publish-pipeline eval_run failure marking are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #409, opened 2026-06-26
refactor: consolidate rate-limit enforcers, fix assert-as-control-flow, pre-compile manifest regex

lfiaschi:
## Context

Ran a deep codebase review (architecture, code health, bugs, performance, testing). This PR ships the **top, high-confidence findings** with verified fixes. Other findings the review surfaced (e.g. potential races in the GitHub crawler, broad excepts in tracker_service, etc.) were either lower-leverage or needed deeper design discussion and are intentionally out of scope here.

### 1. DRY — collapse 9 duplicated rate-limit enforcer functions (`-86` LOC)

Every endpoint that wanted rate limiting was hand-rolling the same 8-line `hasattr / settings / RateLimiter / setattr` dance:

- 7 copies in `server/src/decision_hub/api/registry_routes.py`
- 1 in `server/src/decision_hub/api/auth_routes.py`
- 1 in `server/src/decision_hub/api/search_routes.py`

Extracted into `enforce_rate_limit(name, max_attr, window_attr)` in `rate_limit.py`. The 9 sites now each fit on one line.

While there, **fixed a real race**: under the FastAPI sync threadpool, two threads could both miss the `hasattr` check on cold start and create separate `RateLimiter` instances, the second silently overwriting the first's per-IP tracking state. The factory uses a coarse module-level `threading.Lock` with re-check inside the critical section — no lock cost on the steady-state hot path, race-free on first init.

### 2. Replace `assert` with explicit error in `tracker_routes.py:265`

```python
updated = find_skill_tracker(conn, tid)
assert updated is not None   # ← stripped under `python -O`
return _tracker_to_response(updated)
```

This was the **only `assert`** in the server source tree. Assertions are stripped under `python -O`; if the row ever disappeared between the in-transaction UPDATE and the follow-up SELECT (impossible in practice — same transaction — but still) the code would crash with an `AttributeError` in `_tracker_to_response` instead of returning a clean 500. Replaced with `raise HTTPException(500, ...)`. A regression test pins the behavior.

### 3. Pre-compile YAML key/value regex in `shared/src/dhub_core/manifest.py`

`_load_yaml_with_fallback` was calling `re.match(r"^(\s*[\w][\w-]*:\s*)(.+)$", line)` inside a per-line loop. CPython caches compiled patterns, but the surrounding module already pre-compiles its other patterns at module level — this one was the outlier. Hoisted for consistency and clarity.

### 4. Test coverage gaps closed

- **New `shared/tests/test_taxonomy.py`** — previously **zero** coverage for the canonical category taxonomy (used by classifier, search, CLI, frontend). New tests verify the derived collections (`ALL_SUBCATEGORIES`, `SUBCATEGORY_TO_GROUP`), that `DEFAULT_CATEGORY` is a known subcategory, that subcategories are unique across groups, and that `SkillClassification` is frozen — so an accidental edit to the taxonomy dict that breaks invariants gets caught in CI.
- **New tests in `test_rate_limit.py`** for `get_or_create_limiter` and `enforce_rate_limit` — covering lazy-init, name-scoped caching, and settings lookup. The factory contract is now under test.
- **New `test_update_tracker_disappears_after_update_returns_500`** — regression test for the assert → HTTPException change.

## What was deliberately left out

The audit also surfaced a number of findings I chose not to tackle here because they need more discussion or deeper work:

- Broad `except Exception` blocks in `tracker_service.py` that fail-open on background sync — partly intentional (don't block the critical path) but the error handling could be more nuanced.
- Possible N+1 patterns in `batch_update_github_*` calls — needs profiling on real data before refactoring.
- Embedding storage swallows errors silently in `embeddings.py` — fail-open is intentional today, but a dead-letter retry mechanism would be the right long-term answer.
- Several "missing nil check" findings turned out to be already-handled or pydantic-validated.

A couple of agent-generated findings turned out to be incorrect on inspection and were dropped (e.g. a claimed `dhub_core.ziputil` import typo — it's the correct path per `client/pyproject.toml`; a claimed "missing commit" in the tracker update endpoint — `get_connection` uses `engine.begin()` which autocommits).

## Test plan

- [x] `uv run --package dhub-core --extra dev pytest shared/tests/` — **106 passed** (including 8 new taxonomy tests)
- [x] `uv run --package decision-hub-server --extra dev pytest tests/test_api/` — **233 passed** (including new rate-limit and tracker tests)
- [x] `uv run --package decision-hub-server --extra dev pytest tests/test_domain/` — **450 passed**

https://claude.ai/code/session_017ZJ7nydjiz5rTRPpr7ukHS

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZJ7nydjiz5rTRPpr7ukHS)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Rate-limiter init race fixed more completely by retained #447 (eager registry). Branch preserved.

PR #412, opened 2026-06-29
refactor: tighten eval auth, bound rate limiter, surface tracker errors

lfiaschi:
## What changed
Bundle of small high-leverage fixes from a deep-dive code review: tightens eval-runs authorization, fixes a quadratic per-request hot path and unbounded memory in the in-memory rate limiter, dedupes ~100 lines of copy-pasted rate-limit dependency factories across three route modules, and makes tracker partial-failure errors visible to operators.

## Why
A read-through of `server/src/decision_hub/` surfaced four real issues worth bundling into one focused PR:

1. **`/v1/eval-runs?version_id=…` filtered users in Python.** `find_eval_runs_for_version` fetched every user's runs for a version and then the route did `[r for r in runs if r.user_id == current_user.id]`. That's both slower (full result set leaves Postgres) and a weaker authorization boundary than pushing the filter into the WHERE clause. One bad future refactor and another user's runs leak.
2. **`RateLimiter.__call__` summed every tracked list on every request.** `sum(len(v) for v in self._requests.values())` is O(N-IPs) in the hot path just to decide whether to purge — and the dict was unbounded between purges. A unique-IP flood could grow it without limit.
3. **Nine near-identical `_enforce_*_rate_limit` helpers.** Each function copies the same `hasattr → init → call` pattern. Adding a new rate-limited endpoint meant another copy-paste; changing the pattern meant editing it in nine places.
4. **Tracker partial failures were silently dropped.** When some skills in a repo published successfully and others failed (gauntlet rejection, S3 timeout), the tracker code set `last_error=None`. Operators looking at the tracker saw a "successful" sync with no hint that two of five skills broke.

No linked issue — surfaced via review.

## How to test
- `make test-server` — full server suite (937 tests) passes, including:
  - `tests/test_api/test_eval_logs.py::TestListEvalRuns::test_list_by_version_pushes_user_filter_to_db` — asserts the route passes `user_id` to the DB call (the actual auth check)
  - `tests/test_api/test_rate_limit.py::TestRateLimiter::test_memory_is_bounded_under_unique_ip_flood` — temporarily shrinks `_MAX_TRACKED_IPS` and verifies the dict stays bounded under 500 unique IPs
  - `tests/test_api/test_rate_limit.py::TestRateLimiter::test_call_count_not_scanned_per_request` — guards against re-introducing an O(N) per-request scan
  - `tests/test_api/test_rate_limit.py::TestMakeRateLimitDep` — verifies the factory gives each endpoint an isolated limiter cached on `app.state`
  - `tests/test_domain/test_tracker_service.py::…::test_partial_success_advances_sha_and_surfaces_error` and the 5-skill variant — assert partial failures now reach `last_error`
- `make lint` and `uvx mypy` on the changed modules both pass.
- Manual: hit `/v1/eval-runs?version_id=<some-version>` as user A; previously it would have returned user B's runs for that version after filtering, now the WHERE clause prevents them from being fetched.

## Checklist
- [x] Tests pass (`make test-server`: 937 passed, 34 deselected)
- [x] No breaking API changes — `find_eval_runs_for_version` adds an optional kwarg; route response shape is unchanged
- [ ] Database migration included (if schema changed) — N/A, no schema change

## Notes on what was *not* included
The review surfaced more findings; the ones below were intentionally left out for a future pass:
- Several reviewer claims didn't hold up under inspection (Gemini retry already jitters, the SPA fallback uses a literal allowlist, the search-log background task already uses its own engine).
- Some flagged items are product decisions, not bugs (whether org members or only admins should be able to `publish`).
- A `download_skill` streaming/redirect change would alter CLI behavior and deserves its own focused PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BKnJPMbe7DHNaJVbTwP5Pp)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Cross-tenant eval-run SQL pushdown also in retained lineage; catalogued in #451. Branch preserved.

PR #413, opened 2026-06-30
Harden publish/ZIP pipeline, bound rate-limiter, fix UX retry bugs

lfiaschi:
## What changed

Surgical safety + correctness pass surfaced by a deep multi-agent code review of the server API, database/infra, domain, frontend/CLI, testing, and security layers. 14 files, ~362 net additions across 6 server-side hardenings and 3 client-side bug fixes — each one is small, individually testable, and covered by a new or extended unit test.

## Why

No tracking issue — this PR is the first phase of a broader refactor plan; subsequent phases (splitting `database.py`, `tracker_service.py`, and `gauntlet.py`; adding end-to-end coverage of the publish→gauntlet→eval pipeline) are intentionally left for follow-up PRs to keep this one easy to review.

### Server / domain

- **`shared/dhub_core/ziputil.py`** — `validate_zip_entries()` now rejects symlink entries. Even when an entry name stays inside the target directory, a Unix symlink can point anywhere on the host and would be silently followed at read time.
- **`server/domain/publish.py`** — `extract_for_evaluation()` now (a) rejects negative `file_size` headers up-front (a crafted zip with bogus negatives could otherwise wrap the `sum()` and bypass the 100 MB cap), (b) rejects symlink entries, and (c) decodes scannable files with `errors="strict"`. The old permissive decode silently substituted `U+FFFD` on bad bytes, meaning the gauntlet could scan a sanitised string that no longer matched the bytes that downstream code ends up executing.
- **`server/api/registry_routes.py`** — the download response now uses an RFC 5987 `Content-Disposition` header (quoted ASCII fallback + `filename*=UTF-8''…`). The previous `f'attachment; filename="{...}"'` let user-controlled org/skill names inject quotes, semicolons, or CR/LF into the header.
- **`server/api/registry_routes.py`** — `get_similar_skills()` now passes `user_org_ids=None` to `find_skill_by_slug()` explicitly. No behaviour change today (the parameter defaults to `None`), but it locks the public-only contract so a future refactor cannot quietly widen visibility on this endpoint.
- **`server/api/rate_limit.py`** — `RateLimiter` now caps its tracked-key dict at `_MAX_TRACKED_KEYS=10_000` with a stale-purge plus oldest-key eviction backstop. The previous "every 100 requests" purge could be defeated by a slow spray of unique IPs (especially IPv6) and grow without bound in long-lived containers.
- **`server/infra/database.py`** — wrap the engine-`connect` `SET search_path` cursor in `try/finally` so a transient failure no longer leaks cursors for the lifetime of every pooled connection.

### Frontend / CLI

- **`frontend/components/AskModal.tsx`** — reset `messages` / `query` / `error` when the modal closes so reopening doesn't surface the previous session, and add a Retry button on the error banner so a transient LLM failure no longer forces the user to close-and-reopen the modal.
- **`frontend/pages/SkillDetailPage.tsx`** — the zip-download retry path now explicitly invokes `loadZip()` from the backoff timer (via a ref to avoid stale closures). The previous code relied on a state change re-triggering the dependent `useEffect`, which was brittle and one rename away from silently breaking.
- **`client/cli/runtime.py`** — catch `CalledProcessError` and `FileNotFoundError` from the `uv sync` / `uv run` subprocess calls so the user sees a friendly `typer.Exit` instead of a Python traceback (which could leak the env dict if `pretty_exceptions` are ever enabled).

### Tests
- `shared/tests/test_ziputil.py` — covers symlink rejection (Unix mode bits in `external_attr`).
- `server/tests/test_domain/test_publish.py` — covers invalid-UTF-8, negative-size, and symlink rejection in `extract_for_evaluation`.
- `server/tests/test_api/test_rate_limit.py` — covers tracked-keys cap under high-cardinality IP spray.
- `frontend/components/AskModal.test.tsx` — covers state reset on close/reopen.

## How to test

```bash
# Server (~36s)
uv run --package decision-hub-server --extra dev pytest server/tests/ -q -m "not slow" --ignore=server/tests/test_domain/test_arxiv_gauntlet.py

# Client (~11s)
uv run --package dhub-cli --extra dev pytest client/tests/ -q

# Shared (~1s)
uv run --package dhub-core --extra dev pytest shared/tests/ -q

# Frontend (~6s)
cd frontend && npx vitest run

# Lint + types
uvx ruff@0.15.3 check . && uvx ruff@0.15.3 format --check .
uvx mypy client/src/ server/src/ shared/src/
```

Verified locally — **932 server + 291 CLI + 9 shared + 82 frontend tests pass; ruff and mypy clean.**

## Checklist
- [x] No breaking API changes

---
_Generated by [Claude Code](https://claude.ai/code/session_012rkLSZFYRwtHNDBXnUPgsP)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Zip hardening superseded by retained #438 and the stronger zip-bomb fix from #407 catalogued in #451. Branch preserved.

PR #415, opened 2026-07-02
Harden security, fix races, cut duplication

lfiaschi:
## What changed

Result of a deep pass over server / client / frontend. Fixes real bugs surfaced by the review and collapses duplication, without any behavioural change users can see.

**Server — security / DOS / memory**
- `RateLimiter` now resolves the caller IP from `X-Forwarded-For` via a configurable `trusted_proxy_count` (default `0`; set to `1` for Modal). Behind the LB every request previously shared one bucket, so all `/v1/*`, `/auth/*`, `/ask` limits were effectively global.
- Replaced the fragile `sum(len(v) for v in requests.values()) % 100 == 0` purge (O(N) on every request; could miss the modulo entirely as unique IPs grew, leaking memory) with an interval-throttled purge and deque-based per-key storage for O(1) expiry.
- New `rate_limit_dependency(name)` factory collapses ~10 near-identical `_enforce_*_rate_limit` helpers across `registry_routes`, `search_routes`, `auth_routes` into one thread-safe lazy attach onto `app.state`, using double-checked locking so concurrent first-hits can't lose request history.
- Bounded input sizes on public POSTs (`/v1/keys`, `/auth/github/token`) per the DOS-protection rule in CLAUDE.md.

**Client — credential leaks / permissions / reinstall data loss**
- `core/git_repo.clone_repo`: bounded timeouts, catch `TimeoutExpired`, and sanitise credentials out of git stderr before echoing it back. Previously a tokenised URL like `https://ghp_xxx@github.com/foo` landed verbatim in console + logs on any failure.
- `cli/config.save_config`: write via `0o600` temp-file + atomic `os.replace` so the token file is not world-readable on shared hosts, and a killed process can't leave a half-written config that `load_config` will refuse.
- `cli/registry` install: `rmtree` the skill directory before `extractall` so files removed between versions actually leave. `mkdir(exist_ok=True)` + `extract` was leaving orphaned files from prior installs.

**Frontend — races / design-system rule**
- `useApi`: single monotonic `fetchIdRef` guards both the deps effect and `refetch()`, so a slow `refetch` response can no longer overwrite a fresher one. The old code only guarded the deps effect.
- `SkillDetailPage`: `fetchIdRef` guard on `loadZip` so navigating skill A → B mid-download can't render A's `SKILL.md` and file list on B's page.
- `AskModal`: reset conversation state when `isOpen` goes false so reopening starts clean instead of showing the prior chat.
- Replaced hardcoded `rem` font sizes in `NotFoundPage`, `OrgDetailPage` error branches, `FileBrowser`, `GradeBadge.module.css` with the `--text-*` scale variables (per the CLAUDE.md rule).

## Why

Findings came from a scheduled review pass over the whole codebase (architecture / bugs / security / perf / tests / DX). These are the highest-leverage fixes with the smallest blast radius; larger refactors (splitting `database.py` / `registry_routes.py`, extracting `domain/ask.py`) were noted but intentionally deferred to keep this PR reviewable.

No linked issue — surfaced by the review; happy to file one if useful.

## How to test

- `make lint` (`ruff check .` + `ruff format --check .`) — clean.
- `make typecheck` (`mypy client/src server/src shared/src`) — clean.
- `make test-server` — `942 passed, 34 deselected` including 14 new rate-limit tests (`TestTrustedProxy`, `TestPurge`, `TestFactory`).
- `make test-client` — `301 passed, 29 skipped`, including the new git-subprocess credential / timeout tests, config `0o600` + atomic-save tests, and stale-file reinstall regression test.
- `make test-shared` — `98 passed`.
- `make test-frontend` — `84 passed`, including the new AskModal reset-on-close test and the useApi stale-response guard test.

To verify manually behind a proxy: set `TRUSTED_PROXY_COUNT=1` in `server/.env.dev`, redeploy, and confirm the `X-Forwarded-For` right-most entry is now the per-IP bucket key (e.g. by triggering `/v1/ask` from two different real clients through the same LB and checking that only the offender is 429'd).

## Checklist
- [x] Tests pass (`make test`) — client / server / shared / frontend all green locally.
- [x] No breaking API changes — request/response shapes unchanged; only added optional `trusted_proxy_count` setting (default 0 = old behaviour).
- [x] Database migration included (if schema changed) — n/a, no schema changes.

https://claude.ai/code/session_01SeZmP89R6xuKE7mX3iYZAp

---
_Generated by [Claude Code](https://claude.ai/code/session_01SeZmP89R6xuKE7mX3iYZAp)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its trusted_proxy_count limiter design is superseded by retained #443; frontend race guards overlap retained #446. Branch preserved.

PR #416, opened 2026-07-03
refactor: harden API bugs, dedup boilerplate, plug security + observability gaps

lfiaschi:
## What changed

Deep review of the codebase across API, domain, infra, CLI, and frontend surfaced ~50 findings. This PR lands the highest-leverage subset — a mix of concrete bug fixes, dedup, security hardening, and observability improvements — that fits one review cycle.

**Diff scope:** 22 files, +942 / -326 (18 source files + 4 test files, of which 2 are new).

### Bug fixes

- **`registry_routes` — guarded `json.loads` in eval-log reader.** A partially-written S3 chunk (e.g. from a worker OOM-killed mid-write) used to 500 the entire log-polling endpoint. Now it skips the malformed line with a `logger.warning` and continues.
- **`registry_routes._check_zombie` — return the updated `EvalRun` in memory** instead of re-querying the DB. A concurrent purge could return `None` on the re-read and crash `_run_to_response` with `AttributeError`.
- **`registry_routes.get_similar_skills` — respect visibility.** Threaded `user_org_ids` through `find_skill_by_slug`. Org members were being 404'd on their own private skills' Similar panel.
- **`infra/storage` — paginate `list_objects_v2`** in `list_eval_log_chunks` and batch `delete_eval_logs` at the 1000-key S3 limit. Long-running evals with more than 1000 chunks were silently truncated.
- **`infra/anthropic_client` — retry 429/5xx/529 with `Retry-After` honoring** and filter response blocks by `type == "text"`. Bursty judge loads used to record spurious `verdict="error"` rows, and non-text response shapes (tool_use / thinking) raised `IndexError` on `data["content"][0]["text"]`.
- **`infra/database` — align Python + SQL URL normalization.** `_normalize_repo_url` rtrimmed `/` then stripped `.git`, but the SQL version did the opposite, so a stored URL like `https://x/y.git/` never matched the same repo passed as `https://x/y`. The SQL now lives in a shared `_normalize_repo_url_sql` helper, reused in `mark_skills_source_removed` and `clear_source_removed_for_urls`.
- **`infra/database.search_skills_hybrid`** — cap the merged FTS+vector result set at `limit` (was returning up to `2 × limit` when the two branches disagreed) and add `skills.id` tiebreakers to the ORDER BY so ties don't flip between calls or double the LLM token budget.
- **`domain/repo_utils`** — scrub the GitHub App token from `stdout` too (not just `stderr`) before raising `CalledProcessError`. Some `git` error paths echo the resolved URL, which contains `x-access-token:<token>`, into logs and error trackers.

### Security

- **`keys_routes.StoreKeyRequest`** — bound `value` to 8192 chars and restrict `key_name` to env-var-style `^[A-Z][A-Z0-9_]*$`. The unbounded value was a DoS vector for `user_api_keys`, and lowercase names matched none of the agent key env vars at eval time anyway (`ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, etc.).
- **`cli/config.save_config`** — `chmod` `~/.dhub/config.{env}.json` to `0600` and the parent dir to `0700`. On a shared workstation or CI runner the bearer token used to be world-readable under the default umask.
- **`SkillDetailPage`** — no longer auto-passes `allow_risky=true` for grade-C skills. Users now click through a consent card before the risky zip is fetched, matching the CLI's `--allow-risky` gate. Also fixes the `SkillFile.size` field to be the true byte length (`Uint8Array.byteLength`) rather than the UTF-16 char count of the decoded string — binary files were being reported with the wrong size.

### Performance / reliability

- **`auth_routes.exchange_token`** — moved `sync_org_github_metadata` to `BackgroundTasks`. Login no longer blocks on the 30+ GitHub REST calls that happen the first time each org syncs each day; the JWT returns immediately and avatars/descriptions fill in a moment later.
- **`domain/tracker.check_repo_accessible`** — shortened the default `httpx` timeout to 3s so `POST /v1/trackers` doesn't stall for 10s during GitHub incidents. The check only informs a soft warning.
- **`registry_routes`** — added rate limiters to `/stats` and `/summary` (previously unrated public read paths) and short-lived Cache-Control on `/summary`.

### Dedup / observability

- **`rate_limit.rate_limit_dep(name, max, window)` factory** replaces 8 near-identical `_enforce_*_rate_limit` copies across `registry_routes`, `search_routes`, and `auth_routes`.
- **`deps.parse_uuid`** — shared, replacing the two identical `_parse_uuid` helpers that lived in `registry_routes` and `tracker_routes`.
- **`registry_routes` — two eval-report endpoints (query-string and path-based)** now share a `_fetch_eval_report_or_404` body. Both are actively called (frontend uses `?semver=`, CLI uses the path form), so neither can be deleted, but they no longer drift.
- **`search_routes`** — log at INFO when Gemini references a skill that was never in the retrieval candidates. Previously silently dropped; regressions in the referenced-skills output are now grep-able.
- **`evals` / `registry_service` / `auth_routes`** — replaced `logger.error("... {}", e)` with `logger.opt(exception=True).error(...)` in five places. The plain formatted exception was usually not enough to diagnose sandbox / Modal / GitHub failures; tracebacks reach the Modal logs now.

## Why

The review turned up a handful of correctness bugs (`json.loads` unguarded, zombie re-read race, `.git/` normalization drift), one real security regression on the frontend (drive-by auto-download of risky skills), and a lot of accumulated boilerplate (8 rate-limiter copies, dup `_parse_uuid`, two identical eval-report endpoints). Landing them together is cheaper than filing 15 issues.

## How to test

- `make test-server` — 958 tests pass (up from 942), 34 slow tests deselected.
- `make test-client` — 291 tests pass, 29 skipped.
- `cd frontend && npx vitest run` — 82 tests pass.
- `make lint` and `uvx mypy client/src/ server/src/ shared/src/` — clean.

New coverage worth calling out:
- `server/tests/test_api/test_seo_routes.py` — sitemap.xml XML shape + cache short-circuit, robots.txt host gating (non-prod → `Disallow: /`).
- `server/tests/test_infra/test_normalize_repo_url.py` — freezes the Python↔SQL alignment invariant with the `.git/` regression case.
- `server/tests/test_api/test_registry_routes.py::TestGetSimilarSkills` — auth passes `user_org_ids`, 404 for hidden skill.
- `server/tests/test_infra/test_anthropic_client.py` — retry-then-succeed, no-text-block graceful error, block filter.
- `server/tests/test_infra/test_storage_eval_logs.py` — paginated listing, 1000-key batched delete.
- `server/tests/test_api/test_keys_routes.py` — rejects lowercase key name and oversized value.

## Checklist
- [x] Tests pass (`make test-server`, `make test-client`, `npx vitest run`)
- [x] No breaking API changes for existing clients (both eval-report endpoints preserved; `StoreKeyRequest` rejects only key names that couldn't match an agent env var anyway)
- [x] No schema changes — no migration needed

---
_Generated by [Claude Code](https://claude.ai/code/session_019seGN2G7b6srpt9UynY9H6)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -82,14 +83,25 @@ def load_config() -> CliConfig:
 def save_config(config: CliConfig) -> None:
     """Save CLI config to ~/.dhub/config.{env}.json.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    The config file stores the bearer token. On multi-user machines the
+    default umask would leave it world-readable (0644), so both the
+    directory and the file are locked down to the current user (0700 /
+    0600). ``os.chmod`` is a no-op on Windows filesystems that don't
+    support POSIX modes.
     """
-    CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+    CONFIG_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
+    # If the directory pre-existed with looser permissions, tighten it.
+    # OSError swallowed for filesystems that don't support POSIX modes
+    # (e.g. Windows).
+    with contextlib.suppress(OSError):
+        os.chmod(CONFIG_DIR, 0o700)
     path = config_file()
     path.write_text(
         json.dumps(asdict(config), indent=2) + "\n",
         encoding="utf-8",
     )
+    with contextlib.suppress(OSError):
+        os.chmod(path, 0o600)
 
 
 def get_api_url() -> str:
--- a/frontend/src/pages/SkillDetailPage.tsx
+++ b/frontend/src/pages/SkillDetailPage.tsx
@@ -65,6 +65,11 @@ export default function SkillDetailPage() {
   const [skillMdContent, setSkillMdContent] = useState<string | null>(null);
   const [downloading, setDownloading] = useState(false);
   const [copied, setCopied] = useState(false);
+  // Explicit consent to view/download C-grade (risky) skills. The CLI already
+  // requires `--allow-risky` for the same server endpoint; the browser used to
+  // auto-set this to true when it saw `safety_rating === "C"`, silently
+  // bypassing the risk gate on drive-by visits. Now the user has to click.
+  const [allowRiskyConsent, setAllowRiskyConsent] = useState(false);
   const zipRetries = useRef(0);
   const retryTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
   const MAX_ZIP_ATTEMPTS = 3;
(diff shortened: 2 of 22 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its unique fixes (.git/ normalization drift, 2x-limit search bug, allow_risky consent card) are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #419, opened 2026-07-06
refactor: cross-cutting bug fixes + rate-limit DRY + test hardening

lfiaschi:
## What changed

A cross-cutting review pass across server / CLI / frontend turned up a set of small, high-leverage fixes. This PR bundles the ones I was confident about into a single reviewable diff, backed by tests. See "Detailed rationale" below for the per-file logic.

## Why

Full-repo review (no linked issue) — every fix here is either a real correctness/security bug, a documented CLAUDE.md rule violation, or a copy-paste block worth collapsing. Rejected candidates (non-atomic install, cache concurrency rewrite, broader publish-pipeline surgery) were deliberately kept out of scope.

Closes #

## How to test

- `make lint` — ruff check + format, plus tsc/eslint.
- `make typecheck` — mypy.
- `make test` — full suite: 946 server (`-m "not slow"`) / 295 client / 98 shared / 87 frontend.
- Manual smoke worth eyeballing:
  - `curl -H "X-Forwarded-For: 203.0.113.5, 10.0.0.1" ...` against a rate-limited endpoint — two synthetic "users" no longer share a bucket.
  - `dhub info some/skill --output json | jq .` — flaky sub-request no longer corrupts stdout.
  - Filter change on `/skills` — old cards clear immediately instead of lingering while the new fetch runs.
  - `GET /skills/{org}/{skill}/scan-report?semver=1.2.3` with no scan on 1.2.3 → returns `null` (was: another version's scan).

## Detailed rationale

**Server**
- `api/rate_limit.py` — new `_client_ip()` prefers `X-Forwarded-For` (leftmost) → `X-Real-IP` → transport peer. Behind Modal (or any LB) the transport peer is the same string for every user, so one caller used to 429 everyone else. Also swapped the buggy `total % 100 == 0` purge trigger (fires on cumulative sums, not per N hits) for a plain counter.
- `api/rate_limit.py` — new `make_rate_limit_dependency(name, limit_attr, window_attr)` factory. Collapses 9 hand-rolled `_enforce_*_rate_limit` wrappers in `registry_routes` / `search_routes` / `auth_routes` into one-liners. Zero behaviour change: same `app.state._{name}_rate_limiter` per-container singleton.
- `infra/storage.py` — page through `list_objects_v2` via boto3 paginator in `list_eval_log_chunks` and `delete_eval_logs`. The unpaged calls silently truncated at 1000 keys — long evals lost log chunks past index 1000 and long-lived skills leaked orphan S3 objects on version deletion. `delete_objects` now batches at the 1000-key S3 hard limit.
- `api/registry_routes.py` — `GET /skills/{org}/{skill}/scan-report` no longer substitutes the LATEST scan report for the skill when the caller pinned `?semver=X.Y.Z` and that version has no scan. Users read a substituted scan as "this version was scanned and cleared" — a safety-critical wrong signal. Fallback still runs when semver is omitted (generic skill page).
- `infra/github_client.py` — `graphql()` now goes through the pooled instance `self._client` rather than top-level `httpx.post`. A tracker tick issues many GraphQL calls and the module-level call paid a fresh TCP+TLS handshake each time (~1–2s added per tick on Modal cold containers).
- `infra/cache.py` — `TTLCache` accepts an optional `now` clock so the flaky `time.sleep(0.01)` expiration tests can be rewritten with a fake clock. No production callsite change.

**Client (CLI)**
- `cli/registry.py::_create_zip` — reject symlinks before writing them to the archive and refuse any entry whose resolved path escapes the skill root. Previously `Path.rglob('*')` with `is_file()` returned True for a symlink to a regular file, so a hostile skill with `secrets -> ~/.ssh/id_rsa` would exfiltrate the target's contents into the uploaded zip.
- `core/git_repo.py::clone_repo` — refuse repo URLs / refs starting with `-` and place a `--` separator before the URL in `git clone`. Together they block the `--upload-pack=payload.git` RCE vector (which `looks_like_git_url` would previously accept because it ends in `.git`).
- `cli/config.py::save_config` — create `~/.dhub` at `0o700` and write `config.{env}.json` at `0o600` (opened `O_CREAT|O_WRONLY|O_TRUNC` with mode set at creation so there is no permissive window). Previous behaviour relied on the process umask, which on shared hosts and CI runners produces world-readable bearer tokens.
- `cli/registry.py::info_command` — route the "(could not fetch audit log)" / "(could not fetch eval report)" diagnostics through a stderr `Console`. Previously the dim print landed on stdout unconditionally, corrupting `dhub info --output json` piping when a sub-request flaked.

**Frontend**
- `hooks/useInfiniteScroll.ts` — clear `items` and `total` when deps change so pages using `items.length === 0` as a "first load" signal show the spinner instead of the previous filter's cards under new filter labels. Fixes a 200–800ms visual mismatch on SkillsPage and OrgDetailPage when the user changes org / category / grade.
- `api/client.ts` — percent-encode every dynamic path segment (org slug, skill name, semver) via a shared `encPath` helper. Previously values containing `#`, `?`, `%`, or whitespace produced malformed URLs that routed wrong or 404'd.
- `pages/NotFoundPage.tsx` / `pages/OrgDetailPage.tsx` — replace hardcoded `4rem` / `1.6rem` font sizes on the 404 / error state with the `var(--text-hero)` / `var(--text-heading)` scale variables, per the mobile-first + typography-scale rule in CLAUDE.md.

**Tests**
- `server/tests/test_api/test_rate_limit.py` — new `TestClientIp`, `TestMakeRateLimitDependency`, `TestPurgeCounter` suites, including a regression that two users behind the same proxy do NOT share a bucket.
- `server/tests/test_infra/test_storage_eval_logs.py` — mocks switched to the paginator API and new tests that `list_eval_log_chunks` / `delete_eval_logs` transparently follow >1000 keys.
- `server/tests/test_infra/test_cache.py` — expiration tests rewritten with `_FakeClock`.
- `server/tests/test_infra/test_github_client.py` — assert `graphql()` reuses the pooled client and hits the absolute URL.
- `server/tests/test_api/test_registry_routes.py::TestGetScanReport` — covers both the "pinned semver has no scan → null" fix and the preserved "no pin → fallback still runs" behaviour.
- `client/tests/test_cli/test_registry_cli.py::TestCreateZip::test_skips_symlink_to_file_outside_root` — symlink-escape regression.
- `client/tests/test_core/test_git_repo.py::TestCloneRepoRejectsOptionLikeInputs` — git option-arg regressions.
- `client/tests/test_cli/test_config.py::test_saved_file_is_readable_only_by_owner` — 0o600 regression.
- `frontend/src/api/client.test.ts` — path-encoding regressions for `getSkill` / `resolveSkill` / `getEvalReport` / `downloadSkillZip`.
- `frontend/src/hooks/useInfiniteScroll.test.ts` — new file, asserts items and total are cleared immediately on dep change.

Also: `uv.lock` resync — the lockfile had drifted from client `0.12.3` and shared `0.2.10` already recorded in `pyproject.toml`. `uv run` during the test cycle regenerated it, no version bumps here.

## Checklist
- [x] Tests pass (`make test`) — 946 server / 295 client / 98 shared / 87 frontend
- [x] No breaking API changes — same `/v1/scan-report` shape, `null` return already valid; rate-limiter dependencies are internal

---
_Generated by [Claude Code](https://claude.ai/code/session_01PJjZg2eVk5fTBAFYsjGDcP)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; scan-report semver pinning also fixed in retained #445 lineage and catalogued in #451. Branch preserved.

PR #420, opened 2026-07-07
refactor: fix CLI crashes, dedupe rate limiters, consolidate checksum

lfiaschi:
## What changed

A batch of low-risk correctness fixes, DRY refactors, and observability improvements surfaced by a full-codebase deep-review pass over `client/`, `server/`, `shared/`, and `frontend/`. Every fix ships with a focused regression test.

## Why

The review pass turned up (a) three CLI paths that crash on inputs the server can legitimately produce, (b) a token file written world-readable and non-atomically, (c) eight nearly-identical `_enforce_*_rate_limit` helpers copy-pasted across three route modules, (d) two duplicate `compute_checksum` implementations that could silently drift between client and server, (e) missing tracebacks on the most critical eval-failure log lines, and a few smaller cleanups. None of these are shape-changing enough to warrant separate PRs, but leaving them piecemeal wastes review time.

There's no dedicated GitHub issue for this batch — it's the "action" half of a two-step review-then-refactor task the maintainer asked for.

## How to test

`make test` (already run against a clean checkout: **938 server + 311 client + 105 shared** = 1354 tests pass, plus `make lint` and `make typecheck` clean).

Targeted regressions worth eyeballing:

- **`dhub list`** against a server that emits `"updated_at": null` — should render without crashing.
- **`dhub login`** against a server that emits `"orgs": null` — should complete and save the token.
- **`dhub install --repo git@github.com:acme/foo`** — should reject cleanly instead of mangling the URL into `https://github.com/git@github.com:acme/foo`.
- **`dhub login`** piped into a shell without a TTY — should skip the default-org prompt instead of blocking on stdin.
- **`~/.dhub/config.*.json`** after `dhub login` — file mode should be `0o600`, parent dir `0o700`, and there should never be a `.tmp` sibling lingering after a failed save.
- **`GET /v1/resolve/{org}/{skill}?spec=latest`** with S3 down — download counter must NOT tick up. Pinned by the new `test_download_counter_does_not_advance_on_s3_failure`.
- Rate-limiter dispatch factory: `TestRateLimitDependencyFactory` covers lazy init, independent per-name buckets, and the settings-getter contract.

### Correctness bugs (each with a regression test)

| File | Bug | Failure scenario before fix |
| --- | --- | --- |
| `client/src/dhub/cli/registry.py:659` | `.get("updated_at", "")[:10]` returns `None` when the API sends `"updated_at": null`, then `[:10]` on `None` raises `TypeError`. | One row with a null timestamp aborts the whole `dhub list` render. |
| `client/src/dhub/cli/auth.py:47` | `tuple(token_data.get("orgs", ()))` — `.get(k, ())` still returns `None` when the key is present with `null` value. | Successful login crashes with `TypeError: 'NoneType' is not iterable` and the token is discarded. |
| `client/src/dhub/cli/registry.py:1076` | `_install_from_repo`'s ref parser only checked `startswith("http")`, so an SSH ref `git@github.com:acme/skills` fell into the "prepend `https://github.com/`" branch. | Server 404s with no hint that the URL was mangled. |
| `client/src/dhub/cli/auth.py:78-94` | `_prompt_default_org` (a) blocked on `console.input` even without a TTY (breaks CI logins), and (b) silently defaulted to the first org on empty input despite offering `(none)`. | CI logins hang; interactive users get a surprise default they didn't pick. |
| `client/src/dhub/cli/config.py:82-92` | Token file written with default umask and no atomic rename. | Any user on a shared host reads the bearer token; a killed process leaves a truncated file that surfaces as "Config file is corrupted" on next start. |
| `server/src/decision_hub/api/registry_routes.py:711-737, 744-771` | `increment_skill_downloads` called *before* `generate_presigned_url` / `download_zip_from_s3`. | S3 hiccup → counter still ticks up. Violates the "don't advance state on failure" project rule. |
| `server/src/decision_hub/infra/database.py:2722` | `find_active_eval_runs_for_user` had `ORDER BY created_at DESC LIMIT 10` with no tiebreaker. | Two runs with identical `created_at` render in non-deterministic order (project rule: every `LIMIT` needs a unique tiebreaker). |
| `server/src/decision_hub/infra/cache.py:82-85` | `_evict_one` iterated `self._store.items()` while `del`ing keys — safe today only because of the immediate `return`. | Any future refactor that removes the `return` would `RuntimeError`. |
| `server/src/decision_hub/scripts/backfill_embeddings.py:102` | `logger.info("Backfilled {}/{}", total_processed, total_processed)` printed the same value twice. | Progress log hides actual per-batch progress. |

### Observability

`api/registry_service.run_assessment_background` and three sites in `domain/evals.py` swapped bare `logger.error("... {}", e)` calls for `logger.opt(exception=True).error(...)` so the traceback survives on the failure paths we most need to debug. The two nested error handlers in `run_assessment_background` also stop spinning up fresh SQLAlchemy Engines and reuse the one already in scope.

### DRY refactor: rate-limit factory

`api/rate_limit.py` gains a `rate_limit_dependency(attr_name, settings_getter)` factory. Every route module previously carried a copy of the same 8-line lazy-init block per endpoint (7 in `registry_routes.py`, 1 in `search_routes.py`, 1 in `auth_routes.py`). They now collapse into declarative pairs:

```python
_enforce_publish_rate_limit = rate_limit_dependency(
    "_publish_rate_limiter",
    lambda s: (s.publish_rate_limit, s.publish_rate_window),
)
```

The factory guards the first-time construction with a module-level lock so two threads' concurrent first hits can't build two limiter instances and split the request budget. Behaviour is otherwise byte-identical to the previous copy-pasted implementations. Net: ~60 fewer lines in `registry_routes.py`, ~20 in each of the other two.

### DRY refactor: `_decode_bearer_payload`

`api/deps.py`'s `get_current_user` and `get_current_user_optional` shared ~90% of their body — header parsing, JWT decoding, and the "outdated claim" check. Extracted into a shared `_decode_bearer_payload` helper that returns either a `User` or a sentinel `_AuthReject` rejection. The required-auth variant raises HTTP 401 on rejection; the optional variant logs at DEBUG and returns `None`.

### Cross-package: `dhub_core.hashing`

New `shared/src/dhub_core/hashing.py` provides `sha256_hex` and `verify_sha256` as the single source of truth for the publish checksum. Both `client/src/dhub/core/install.py:compute_checksum` and `server/src/decision_hub/infra/storage.py:compute_checksum` now delegate to it. Previously the two copies could silently drift, breaking the "no-op publish" detection that compares client-side checksums against the server's stored checksum.

### Cleanups

- `_SKILL_SUMMARY_KEYS` in `infra/database.py:700` drops the `hasattr(col, "key") else col.name` fallback — every entry in `_SKILL_SUMMARY_COLUMNS` has `.key`, so the else branch was dead.

### Tests added

- `shared/tests/test_hashing.py` — 7 tests pinning the canonical checksum contract.
- `client/tests/test_cli/test_render.py` — 3 tests around `_render_skills_table` null-safety.
- `client/tests/test_cli/test_auth_prompt.py` — 9 tests around `_prompt_default_org` (TTY, non-TTY, empty input, unknown org, Ctrl-D).
- `client/tests/test_cli/test_install_ref_parsing.py` — 4 tests around `--repo` ref normalisation.
- `client/tests/test_cli/test_auth_cli.py` — 1 new test for `"orgs": null` in the login response.
- `client/tests/test_cli/test_config.py` — 4 new tests around atomic writes and 0o600 file mode.
- `server/tests/test_api/test_rate_limit.py` — 4 tests for the new dependency factory.
- `server/tests/test_api/test_registry_routes.py` — 1 test pinning the counter-after-S3 ordering for `/v1/resolve`.

## Checklist

- [x] Tests pass (`make test`) — 1354 pass, 0 fail.
- [x] Lint clean (`make lint`), typecheck clean (`make typecheck`).
- [x] No breaking API changes — no request/response schemas or endpoint URLs changed.
- [x] Database migration included — none needed, no schema change.

## Deferred (not in this PR)

The review turned up several higher-effort follow-ups that would need their own PRs and their own dedicated review time — most notably splitting `server/infra/database.py` (3465 lines) and `client/cli/registry.py` (1912 lines) into cohesive submodules, and pushing 66 domain-layer `logger` calls back into the API/infra callers per the CLAUDE.md rule. These are documented in the review scratchpad but intentionally out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XLrMqUSmSfEcXkSXbYNk5i)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its shared hashing module and CLI login fixes are catalogued in #451. Branch preserved.

PR #422, opened 2026-07-09
core hardening: publish path, rate limiting, and correctness

lfiaschi:
## What changed

Full-tree principal-engineer review of the monorepo followed by a coherent
hardening pass. This PR ships the highest-leverage subset of the review's
findings — real bugs, real duplication, real security holes — with tests.
The rest of the review is captured below so follow-ups can be triaged.

## Why

Standing routine sweep of the codebase for correctness, security, and
simplification wins. No linked issue; findings are self-contained.

Closes #

## How to test

- `make test` from the repo root (Python and frontend).
- `make lint` (ruff + frontend).
- Locally: verified 234 API + 620 domain/infra + 300 client + 102 shared +
  82 frontend tests pass, ruff clean, format clean, TypeScript clean.

## Checklist
- [x] No breaking API changes (public HTTP contract unchanged; two new
      opt-in things: `Settings.trust_forwarded_for` default True, and
      `find_eval_runs_for_version` gained an optional `user_id=` kwarg).
- [x] Database migration included (n/a — no schema changes)

---

# Highlights (what this PR fixes)

### Server / rate limiting
- Collapse 9 near-identical `_enforce_*_rate_limit` helpers (across
  `registry_routes`, `auth_routes`, `search_routes`) into a single
  `rate_limit(name)` factory in `api/rate_limit.py`. Sites become
  `dependencies=[limit("publish")]`.
- **Bug fix**: behind Modal's LB, `request.client.host` is the LB IP, so
  every real caller shared one bucket — the limiter was effectively off,
  or (worse) 429'd everyone the moment the LB IP crossed the threshold.
  Introduced `Settings.trust_forwarded_for` (default True) and now key
  buckets by the left-most `X-Forwarded-For` token; keeps the door shut
  for local dev where the header is user-controllable
  (`trust_forwarded_for=False`).
- Purge cadence switched from `sum(len(v) for v in self._requests.values()) % 100`
  (O(N tracked IPs) *every call*) to a monotonic `_call_count % _PURGE_INTERVAL`.

### Server / publish path
- New `parse_skill_md_content(content)` in `dhub_core.manifest`; the
  server's `parse_manifest_from_content` no longer round-trips through
  `tempfile.NamedTemporaryFile` + write + read + unlink on every publish.
- `parse_skill_md` now opens files as `utf-8` (default `cp1252` on
  Windows fails on any non-ASCII character in the description).

### Server / eval-run correctness + perf
- `list_eval_runs?version_id=...` used to fetch *every* user's runs and
  filter in Python. For a popular public skill that moves the entire set
  across the wire before discarding all but the current user's rows.
  Pushed the filter into SQL via a new `user_id=` kwarg on
  `find_eval_runs_for_version`. Test updated to lock the DB call receives it.
- `_check_zombie` now returns the mutated `EvalRun` (via
  `dataclasses.replace`) so `get_eval_run` no longer runs a redundant
  `find_eval_run` immediately after its own UPDATE.

### CLI / publish security
- `_create_zip` used to silently follow symlinks. A hostile skill
  directory could ship `id_rsa -> ~/.ssh/id_rsa` and the server would
  package the key as the skill payload. Now refuses any file symlink AND
  any file whose ancestor directory is a symlink; also verifies the
  resolved path stays inside the skill root.
- `discover_skills` used to follow directory symlinks. A cyclic
  `a/link -> ..` would recurse until Python bailed; a symlink pointing
  outside the tree would double-publish skills under misleading paths.
  Switched to `os.walk(followlinks=False)` and pruned hidden dirs in-place.

### CLI / reliability
- `~/.dhub/config.{env}.json` is now written atomically
  (`tempfile.mkstemp` + `os.replace`) and `chmod`'d to `0600` since it
  holds a bearer token. Prevents (a) two concurrent CLI processes
  corrupting the file and forcing `rm ~/.dhub/*.json && dhub login`, and
  (b) the token being world-readable on shared hosts.
- `clone_repo` now bounds every git subprocess with `timeout=180`
  through a shared `_run_git` helper. Previously a slow/hostile remote
  could hang `dhub publish <git-url>` indefinitely with only a spinner
  for feedback.

### Frontend
- `useApi` now sets `data = null` on dep change so callers rendering
  `data` unconditionally don't flash the previous route's payload while
  the new fetch is in flight (per project rule *"reset state on context
  changes"*).
- `SkillDetailPage` retry `setTimeout` (up to ~6s tail on attempt 3) now
  clears on **unmount**, not just on skill change — otherwise it fires
  `setState` on a dead component.
- SKILL.md front-matter strip regex accepts `\r?\n` so a CRLF-authored
  SKILL.md doesn't render the raw YAML block in the Overview tab.

### Tests added / updated
- Rate-limiter: X-Forwarded-For keying, factory, purge cadence, trust flag.
- `_create_zip`: file-symlink and directory-symlink refusal.
- `discover_skills`: cyclic-symlink safe, outside-tree symlink ignored.
- `clone_repo`: timeout path + `timeout=` actually passed.
- Config: `0600` mode, atomic-write leaves no temp cruft, failed write
  leaves the existing config intact.
- `parse_skill_md_content` parity with the file-path variant + utf-8 lock.
- `useApi`: data-reset lock on dep change.
- `test_list_by_version_filters_to_current_user`: asserts DB call
  receives `user_id=`.

---

# Follow-ups from the review (not in this PR)

These were surfaced during the survey; each is a plausible next PR. Kept
separate so this diff stays reviewable.

### Server — moderate value
- `batch_update_github_stars` in `infra/database.py` fires one UPDATE per
  repo on every tracker tick; consolidate into one `UPDATE … FROM UNNEST`
  and escape the LIKE wildcards (`repo_url` containing `_` matches other
  repos).
- Gemini retry helper in `infra/gemini.py` catches only `TimeoutException`;
  a transient TCP reset (`ConnectError`, `ReadError`) propagates and
  rejects the publish. Widen to `httpx.RequestError` and hoist client
  creation outside the retry loop.
- `get_scan_report` silently falls back to a scan report for a
  *different* version when the requested one has none — the response has
  no explicit `is_stale` flag; add one or 404 properly.
- `list_eval_log_chunks` re-fetches every S3 chunk on every poll because
  the cursor is an *event* seq and the S3 list uses a *chunk* seq. Store
  the chunk's `[first_seq, last_seq]` in a per-run index so subsequent
  polls can skip finished chunks.
- Publish-pipeline `_build_*_fn` factories duplicate a ~20-line closure
  five times (analyze / analyze-prompt / review-body / review-code /
  analyze-credential). Collapse into a single generic factory.
- `resurrect_removed_skills` returns overlapping keys `resurrected` +
  `skills_resurrected` at different granularities; drop one.
- `update_org_github_metadata` clobbers unpassed columns with NULL; use
  the sentinel pattern from `update_skill_tracker`.
- Rate limits are keyed lazily via LB IP; consider running the app
  behind `ProxyHeadersMiddleware` with an explicit `trusted_hosts`
  allowlist rather than blanket-trusting the header.

### CLI — moderate value
- Multiple `console.print(...)` calls in `_auto_detect_org`,
  `_publish_from_git_repo`, and `search.ask_command` violate JSON mode
  (breaks downstream JSON parsers).
- Bare `except Exception:` around auto-tracker creation swallows errors
  silently — users can't diagnose why their repo isn't tracked.
- `_render_skills_table` and `_render_skill_info` do not escape Rich
  markup in `description`, `author`, `category`. A skill with
  `[red on white]hijacked[/]` in its description renders styled text.
- `.dhub-version` and PyPI version cache use non-atomic `write_text` too;
  minor but easy to unify with the new atomic-write helper.
- `_looks_like_sha` matches `deadbeef` and similar words → unnecessarily
  does a full clone.

### Frontend
- All fetches lack `AbortController` — rapid filter/search changes on
  `SkillsPage`/`OrgsPage` fire N in-flight requests. Thread `signal`
  through `fetchJSON` and `useApi`'s auto-fetch effect.
- Ad-hoc `window.dispatchEvent(new CustomEvent("open-ask-modal"))` used
  as a cross-component bus; replace with a `AskModalContext`.
- Duplicated `IntersectionObserver` sentinel logic in `OrgsPage` and
  `useInfiniteScroll` — extract a `useSentinelObserver` hook.
- Hardcoded `rem` font sizes in `NotFoundPage`, `OrgDetailPage`,
  `FileBrowser`, `GradeBadge.module.css` violate the fixed `--text-*`
  scale rule from `CLAUDE.md`.
- `FileBrowser`'s `buildTree(files)` runs on every render — wrap in
  `useMemo(_, [files])`.
- Feature flags `LINK_TO_MANIFEST` and `SHOW_SCANNER_REPORT` are
  hardcoded `false` — either promote to env-driven or delete the dead
  ~334-line `ScannerReport.tsx` code path.

### Test coverage gaps
- Whole modules with zero tests: CLI `app.py` (`upgrade_command`,
  `_detect_installer`), CLI `banner.py`, CLI `config_cmd.py`.
- Frontend: `OrgDetailPage`, `Layout`, `FileBrowser`, `useInfiniteScroll`
  have no tests.
- No dedicated DB-level test file for eval-run helpers (added contract
  assertion in the route test but a direct DB-level test would be
  stronger).
- Client and shared duplicate ~40 manifest tests through the
  re-export shim; a single re-export test would suffice.

https://claude.ai/code/session_01EtNvS42VmBuP7bf7XvaE1D

---
_Generated by [Claude Code](https://claude.ai/code/session_01EtNvS42VmBuP7bf7XvaE1D)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Broad superset sweep; its discover_skills symlink traversal and parse_skill_md_content fixes are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #424, opened 2026-07-12
fix: harden search, evals and external clients (deep-review sweep)

lfiaschi:
## What changed

Deep-review sweep of the server package. Eleven low-risk, high-confidence fixes across SQL correctness, external-client reliability, cache safety, and eval-run consistency — plus five new regression tests. No public API surface changes; every change is a bug fix or hardening on a stable contract.

## Why

Standalone architectural / correctness review, not a specific issue. Findings that made this cut all met the bar of: concrete failing scenario, small blast radius, clear fix.

## How to test

`make test-server` — all 944 non-slow server tests pass, `make lint` clean, `mypy` clean on the touched modules.

Manual verification is not required for any of these — every fix is covered by an existing or new unit test.

## Checklist

- [x] Tests pass (`make test-server`, 944 passed / 34 slow deselected)
- [x] No breaking API changes (all endpoints/signatures unchanged)

---

### SQL correctness — LIMIT queries need a unique tiebreaker

Per the project convention documented in `CLAUDE.md` ("SQL query correctness"), every `LIMIT` query needs an explicit tiebreaker in `ORDER BY`. Without one, two rows with identical rank/distance/timestamp can be silently dropped or duplicated between calls.

- **`search_skills_hybrid`** — `skills.id DESC` added after both `fts_rank DESC` and `vec_dist ASC`. Zero-rank FTS misses and identical embeddings (forks, duplicate content) previously produced non-deterministic top-K across calls.
- **`fetch_similar_skills`** — same tiebreaker, plus a bigger cleanup: the two-query source-embedding lookup + neighbour search collapses into a single CTE. Halves the round-trip on the `/skills/.../similar` hot path.
- **`find_active_eval_runs_for_user`** — parallel `dhub publish` invocations can create runs in the same millisecond; adding `id DESC` keeps `/eval-runs` pagination stable and prevents runs from disappearing/reappearing between polls.
- **`list_tracker_metrics`** — same treatment for the tracker metrics list.

### Correctness — batch counter must not grow unbounded

`batch_increment_permanent_failures`'s docstring says a tracker already at threshold "is not incremented further" — but the UPDATE had no WHERE clause enforcing that. A disabled tracker that got retried would have its counter climb into the thousands. Adds `consecutive_permanent_failures < threshold` to the UPDATE.

### Correctness — zombie detection missing from `list_eval_runs`

`GET /v1/eval-runs/{id}` and `GET /v1/eval-runs/{id}/logs` both call `_check_zombie` before returning, but the collection endpoint `GET /v1/eval-runs` didn't. A crashed worker's run therefore appeared "running" indefinitely in the "My Runs" list. Now applies the same stale-heartbeat check and re-reads the list if any run was updated in-DB.

### Reliability — external clients

- **S3 client timeouts** — `create_s3_client` now uses `botocore.config.Config(connect_timeout=5, read_timeout=30, retries={"max_attempts": 3, "mode": "adaptive"})`. boto3's defaults are 60s per attempt × 5 retries, plenty for the request pool to run out on a slow bucket during a publish or download.
- **Anthropic judge retry** — `judge_eval_output` was a single `httpx.post`; a flaky 429/502 failed the whole gauntlet case. Added exponential-backoff retry that mirrors `gemini_request_with_retry` for 429/500/502/503/529 and `httpx.TimeoutException` (up to 3 retries with jitter).
- **Gemini 403 handling** — 403 was blanket-retriable, but Gemini returns 403 for both quota exhaustion *and* invalid API keys. On a rotated/misconfigured key, every gauntlet call ate ~7s of retry budget before failing (×4 stages ≈ 28s per publish). Now retries 403 only when the response body contains "quota" / "rate" / "resource has been exhausted"; other 403s fail fast.
- **Tracker verification fail-safe** — `_verify_repos_removed` in the tracker service now catches `httpx.HTTPError` per-URL and treats a transient failure as "still exists". A network blip must not silently flip live skills to `source_repo_removed=true`.

### Perf / security — cache-eviction resistance on `/orgs/stats`

`GET /orgs/stats` was caching under `f"org_stats:{search}:{type_filter}:{sort}:{sort_dir}"`, embedding the arbitrary user `search` string in the key. A low-QPS attacker sending random queries could evict genuine hot entries from the bounded 256-entry TTLCache. The route now caches only the fixed `(type_filter, sort, sort_dir)` matrix; requests with a `search=` parameter bypass the cache entirely.

### Tests added

- `test_infra/test_search_db.py` (new file) — asserts compiled SQL for `search_skills_hybrid`, `fetch_similar_skills`, and `find_active_eval_runs_for_user` contains the required tiebreakers, and that `fetch_similar_skills` runs exactly one `execute()` call.
- `test_infra/test_tracker_db.py` — LIMIT tiebreaker on `list_tracker_metrics`; UPDATE bound at threshold for `batch_increment_permanent_failures`.
- `test_infra/test_anthropic_client.py` — retry on 429 then succeed; fast-fail on 400; four total attempts before propagating 5xx.
- `test_infra/test_gemini.py` — 403 with quota body retries; 403 with permission body fails fast (regression for the invalid-key retry-waste behaviour).
- `test_api/test_eval_logs.py` — regression for zombie detection in the `list_eval_runs` endpoint.

---

## What this PR does *not* include (deliberately left out)

The review surfaced several additional candidates that were held back because they were higher-risk or needed a decision I didn't want to make autonomously:

- Auth login handler `async def` refactor (sync SQLAlchemy calls blocking the event loop) — real issue, but semantics change.
- Moving `sync_org_github_metadata` off the login critical path via `BackgroundTasks` — same reason.
- S3 zip download streaming (currently reads the whole zip into memory) — behaviour change on a public endpoint.
- Download counter idempotency / removal from `/resolve` — product decision.
- Rolling back `insert_skill` on late S3 failure in `publish_pipeline` — needs a careful transactional-boundary discussion.
- Modal orphan-labeling correctness under deadline break — needs a broader rework of the fn.map consumer.
- Dead-code removal (`TestingConfig`, `_normalize_repo_url`, various shim re-exports, some frontend feature-flag dead branches) — worth a follow-up PR of its own.

These are documented in the review notes; happy to open follow-ups.

https://claude.ai/code/session_01VMuJD2R4E97Fb4veBXGZYN

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMuJD2R4E97Fb4veBXGZYN)_

### DIFF
--- a/server/src/decision_hub/api/org_routes.py
+++ b/server/src/decision_hub/api/org_routes.py
@@ -167,10 +167,14 @@ def get_org_stats(
     if ttl:
         response.headers["Cache-Control"] = f"public, max-age={ttl}"
 
-    cache_key = f"org_stats:{search}:{type_filter}:{sort}:{sort_dir}"
-    cached = cache.get(cache_key)
-    if cached is not None:
-        return cached
+    # Only cache the fixed filter matrix. Embedding an arbitrary user
+    # `search` string in the cache key lets an attacker sending random
+    # queries evict genuine hot entries from the bounded TTLCache.
+    cache_key = f"org_stats::{type_filter}:{sort}:{sort_dir}" if not search else None
+    if cache_key is not None:
+        cached = cache.get(cache_key)
+        if cached is not None:
+            return cached
 
     rows = fetch_org_stats(conn, search=search, type_filter=type_filter, sort=sort, sort_dir=sort_dir)
     items = [
@@ -185,7 +189,7 @@ def get_org_stats(
         for row in rows
     ]
     result = OrgStatsResponse(items=items)
-    if ttl:
+    if cache_key is not None and ttl:
         cache.set(cache_key, result, ttl=ttl)
     return result
 
--- a/server/src/decision_hub/api/registry_routes.py
+++ b/server/src/decision_hub/api/registry_routes.py
@@ -1194,13 +1194,34 @@ def list_eval_runs(
     conn: Connection = Depends(get_connection),
     current_user: User = Depends(get_current_user),
 ) -> list[EvalRunResponse]:
-    """List eval runs, optionally filtered by version ID."""
+    """List eval runs, optionally filtered by version ID.
+
+    Applies the same zombie-heartbeat detection as the single-run endpoints
+    so a crashed worker's run does not appear "running" indefinitely in the
+    "My Runs" list. Runs whose heartbeat has gone stale are updated in-DB
+    and returned with status="failed".
+    """
     if version_id is not None:
         parsed_vid = _parse_uuid(version_id, "version_id")
         runs = find_eval_runs_for_version(conn, parsed_vid)
         runs = [r for r in runs if r.user_id == current_user.id]
     else:
         runs = find_active_eval_runs_for_user(conn, current_user.id)
+
+    updated_any = False
+    for r in runs:
+        new_status = _check_zombie(conn, r)
+        if new_status != r.status:
+            updated_any = True
+    if updated_any:
+        # Re-read so response reflects the failed status without threading
+        # mutations back through the frozen model.
+        if version_id is not None:
+            runs = find_eval_runs_for_version(conn, parsed_vid)
+            runs = [r for r in runs if r.user_id == current_user.id]
+        else:
+            runs = find_active_eval_runs_for_user(conn, current_user.id)
+
     return [_run_to_response(r) for r in runs]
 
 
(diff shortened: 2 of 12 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its Gemini-403-quota-only retry and orgs/stats cache-eviction fixes are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #425, opened 2026-07-13
refactor(api): DRY rate-limit boilerplate + fix hot-path perf & memory leaks

lfiaschi:
## What changed

Three surgical fixes to the API rate-limiter layer surfaced during a code review pass. All internal — no behaviour visible to CLI clients or the frontend.

- **`rate_limit_dep()` factory** replaces the 9 hand-rolled `_enforce_*_rate_limit` helpers scattered across `auth_routes.py`, `registry_routes.py`, and `search_routes.py`. Each one did the same lazy-init dance around a `RateLimiter`, differing only in the two settings-attribute names. Net −50 lines, and one place to touch when the caching layout changes.
- **`RateLimiter` hot-path is now O(1).** The previous purge trigger was `sum(len(v) for v in self._requests.values()) % 100 == 0` — a full scan of the tracked-IP dict on every request, growing with container traffic. Swapped for a simple counter that fires the purge every `_PURGE_INTERVAL_CALLS` (=100) requests; same effective cadence, no scan.
- **Reject-branch memory leak fixed.** When an IP hit its cap, the pruned timestamp list was only assigned back on the *allowed* branch. A client that kept hammering after being throttled kept accumulating stale entries behind the scenes — the list only kept growing until the periodic purge dropped it whole. Regression test added.
- **`CLIVersionMiddleware`** now early-returns when the header is missing instead of relying on Python short-circuit evaluation to skip an unassigned `client_parsed`. Same semantics, no risk of `NameError` if the two `if client_ver:` guards drift apart.

## Why

Surfaced during an architect/principal-engineer review pass. No linked issue — pure code-health work.

- `_enforce_*_rate_limit` duplication is the kind of thing that grows until someone gets tired and DRYs it. Nine call sites was already the tipping point.
- The `total = sum(...)` line was on the hot path for every rate-limited request (search, publish, download, resolve, list, similar, audit-log, scan-report, auth). Under load with many unique client IPs, that scan is real overhead.
- The reject-branch leak is subtle but real: a stubborn client can only add entries when it's *not* being rate-limited today, so the leak is bounded by fresh traffic — but any IP that flips between throttled and allowed will keep growing until the next 100-call purge. Fixing it costs one assignment.

## How to test

```bash
make test-server                 # 938 tests pass, -m "not slow"
```

New tests in `server/tests/test_api/test_rate_limit.py`:

- `test_rejected_request_still_prunes_stale_entries` — regression for #3 above.
- `test_purge_stale_removes_inactive_ips` — asserts the new counter-based purge trigger.
- `TestRateLimitDep::test_lazily_creates_limiter_on_first_call` — factory contract.
- `TestRateLimitDep::test_reuses_existing_limiter_across_requests` — the cached limiter is reused, not rebuilt.
- `TestRateLimitDep::test_enforces_configured_limit` — end-to-end: settings → limiter → 429.

Also verified:

- `uvx ruff@0.15.3 check` + `format --check` — clean.
- `uvx mypy` on the five touched modules — clean.

## Checklist
- [x] Tests pass (`make test-server`, 938 passed)
- [x] No breaking API changes
- [x] No database migration needed

---
_Generated by [Claude Code](https://claude.ai/code/session_01YTWcxC4TtXc8jWcM9qmFRW)_

### DIFF
--- a/server/src/decision_hub/api/app.py
+++ b/server/src/decision_hub/api/app.py
@@ -96,18 +96,21 @@ class CLIVersionMiddleware:
 
         # Only enforce version check for CLI requests (those sending the header).
         # Browser / frontend requests don't send the header and should pass through.
+        if not client_ver:
+            await self.app(scope, receive, send)
+            return
+
         # Malformed version headers are treated as outdated — return 426 so the
         # client upgrades to a version that sends a valid semver header.
-        if client_ver:
-            try:
-                client_parsed = _parse_semver(client_ver)
-            except ValueError:
-                client_parsed = (0, 0, 0)
-        if client_ver and client_parsed < self._min_parsed:
+        try:
+            client_parsed = _parse_semver(client_ver)
+        except ValueError:
+            client_parsed = (0, 0, 0)
+        if client_parsed < self._min_parsed:
             body = _json.dumps(
                 {
                     "detail": (
-                        f"Your CLI version ({client_ver or 'unknown'}) is below the "
+                        f"Your CLI version ({client_ver}) is below the "
                         f"minimum required ({self.min_version}). "
                         "Run 'uv tool install --upgrade dhub-cli' or "
                         "'pip install --upgrade dhub-cli' to update."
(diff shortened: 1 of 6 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Strict subset of the rate-limiter work in retained #447. Branch preserved.

PR #426, opened 2026-07-14
fix: correctness bugs and CLAUDE.md-rule violations surfaced by a codebase audit

lfiaschi:
## What changed

A read-only audit of the server, CLI, and shared packages surfaced a small set of concrete correctness bugs and CLAUDE.md-rule violations. This PR fixes them, adds regression tests, and leaves behaviour otherwise unchanged.

## Why

No linked issue — surfaced by a codebase-wide review. Each fix maps to a concrete failing scenario (details below).

Closes #

## How to test

`make test lint typecheck` all pass locally. Suite-by-suite:

- shared: `99 passed`
- client: `292 passed, 29 skipped`
- server: `941 passed` (`-m "not slow"`)
- frontend (`npx vitest run`): `82 passed`
- `uvx ruff@0.15.3 check .` and `format --check .` clean
- `uvx mypy client/src/ server/src/ shared/src/`: `no issues found in 88 source files`

Manually verifiable regressions each new test locks in:

1. **`tracker_service`** — `TestDisabledCountPreservedOnEarlyReturn::test_disabled_count_survives_rate_limit_early_return` fails on `main` (returns `trackers_disabled=0`).
2. **`search_routes`** — `test_ask_fallback_analytics_records_actual_answer` fails on `main` (analytics record `answer=""`).
3. **`storage`** — `test_paginates_across_multiple_pages` / `test_deletes_across_multiple_pages` fail on `main` (only first 1000 keys handled).
4. **`org_routes`** — `test_org_stats_returns_429_after_limit` / `test_org_profiles_returns_429_after_limit` fail on `main` (no rate limiter).
5. **CLI** — `test_list_handles_null_updated_at` fails on `main` (`None[:10]` TypeError crashes `dhub list`).
6. **shared/manifest** — `TestParseSkillMdEncoding::test_utf8_content_parses_when_file_is_utf8` fails on `main` when the process locale is not UTF-8 (Windows cp1252).

## Checklist
- [x] No breaking API changes
- [x] Database migration included (if schema changed) — not applicable, no schema changes

---

### Detailed fixes

**Server**

- `domain/tracker_service.py` — the two guardrail early-return paths in `check_all_due_trackers` (GitHub rate-limit low; dispatch deadline too close) hard-coded `trackers_disabled=0`, silently dropping the count of trackers the permanent-failure circuit breaker had already disabled earlier in the same tick. Both branches now propagate `disabled_count`.
- `api/search_routes.py` — the LLM-failure fallback recorded `answer=""` in analytics while the response body returned "Here are the most relevant skills I found:". Extracted the string into a local `fallback_answer` so response and analytics cannot drift.
- `infra/storage.py` — `list_eval_log_chunks` and `delete_eval_logs` used `list_objects_v2` directly, which caps at 1000 keys. Long streaming evals whose `chunk_seq` crossed 1000 silently lost the tail of their log (and `delete_eval_logs` leaked objects). Switched both to the paginator; the delete path batches deletes per page since `delete_objects` also has a 1000-key cap.
- `scripts/backfill_embeddings.py` — added `.order_by(skills_table.c.id)` on the `LIMIT` query to satisfy the SQL-correctness rule from `CLAUDE.md` (a persistently failing batch could otherwise starve later work), and snapshotted the pending count so progress logs read `500/12000` instead of the nonsensical `500/500`.
- Loguru rule (`CLAUDE.md` requires `logger.opt(exception=True)` rather than formatting exceptions into the message string) — fixed six sites across `api/registry_service.py` and `domain/evals.py`. On-call now sees tracebacks in stderr for background eval failures instead of a bare "sandbox error: <msg>" line.
- `api/org_routes.py` + `settings.py` — the public unauthenticated `/v1/orgs/stats` and `/v1/orgs/profiles` endpoints, both fronting expensive aggregate queries, had no rate limiter. Added `_enforce_org_stats_rate_limit` and `_enforce_org_profiles_rate_limit` mirroring the existing `_enforce_*_rate_limit` pattern, with settings (`org_stats_rate_limit` / `org_profiles_rate_limit`, default 120/60s each).

**CLI / shared**

- `client/src/dhub/cli/registry.py` — the Ctrl-C detach message in `publish_command` was a plain string, so users saw the literal token `{eval_run_id}` instead of a copyable id. Added the `f` prefix.
- `client/src/dhub/cli/registry.py` — `_render_skills_table` did `s.get("updated_at", "")[:10]`, which returns `None` when the key is present with a null value; slicing then blew up the entire `dhub list` render with `TypeError`. Guarded with `(s.get("updated_at") or "")[:10]`.
- `shared/src/dhub_core/manifest.py` and `client/src/dhub/cli/init.py` — pinned `encoding="utf-8"` on the SKILL.md read and write so a Windows user whose locale defaults to cp1252 can round-trip em-dashes, curly quotes, and accented characters in the description. Every other file I/O in the project already sets UTF-8 explicitly; these two were the outliers.

**Tests**

- `server/tests/test_domain/test_tracker_service.py` — regression test for the disabled-count fix.
- `server/tests/test_api/test_search_routes.py` — regression test that the fallback analytics record the same `answer` as the response body.
- `server/tests/test_infra/test_storage_eval_logs.py` — rewrote to mock the paginator API; added `>1000 chunks` list-and-delete regression tests.
- `server/tests/test_api/test_org_routes.py` + `conftest.py` — 429 assertions for both new rate limiters.
- `client/tests/test_cli/test_registry_cli.py` — CLI list test with `updated_at=null`.
- `shared/tests/test_manifest.py` — UTF-8 SKILL.md parse test.

### Not in scope for this PR

The following review findings were surfaced but need behaviour discussion before touching:

- `RateLimiter` keys on `request.client.host`, which collapses to Modal's proxy IP. Fix would prefer `X-Forwarded-For` — but requires validating what Modal actually sets.
- `find_active_eval_runs_for_user` is misnamed (it does not filter by status). Renaming touches an established API surface.
- `get_eval_run` does a double-read after `_check_zombie` on every request — trivial to fix but not a bug, and the polling loop is not currently a bottleneck.

https://claude.ai/code/session_01ExQM2wkZaZSayZMmgsHDx5

---
_Generated by [Claude Code](https://claude.ai/code/session_01ExQM2wkZaZSayZMmgsHDx5)_

### DIFF
--- a/client/src/dhub/cli/init.py
+++ b/client/src/dhub/cli/init.py
@@ -38,6 +38,10 @@ def init_command(
         console.print(f"[red]Error: {skill_md} already exists.[/]")
         raise typer.Exit(1)
 
+    # Pin UTF-8 explicitly so a Windows user typing a curly quote or
+    # em-dash in the description does not write cp1252-encoded bytes that
+    # later fail to parse (SKILL.md is read as UTF-8 by both the CLI and
+    # the server).
     skill_md.write_text(
         f"---\n"
         f"name: {name}\n"
@@ -46,7 +50,8 @@ def init_command(
         f"\n"
         f"# {name}\n"
         f"\n"
-        f"Describe what this skill does and how the agent should use it.\n"
+        f"Describe what this skill does and how the agent should use it.\n",
+        encoding="utf-8",
     )
 
     console.print(f"[green]Created skill project at {skill_dir.resolve()}[/]")
--- a/client/src/dhub/cli/registry.py
+++ b/client/src/dhub/cli/registry.py
@@ -155,7 +155,7 @@ def _publish_skill_directory(
 
             _tail_eval_logs(api_url, _bh(token), eval_run_id)
         except KeyboardInterrupt:
-            console.print("\n[dim]Detached. Resume with: dhub logs {eval_run_id} --follow[/]")
+            console.print(f"\n[dim]Detached. Resume with: dhub logs {eval_run_id} --follow[/]")
     elif eval_report_status == "pending":
         console.print("[dim]Agent assessment running in background...[/]")
 
@@ -656,7 +656,11 @@ def _render_skills_table(skills: list[dict], title: str = "Published Skills") ->
     for s in skills:
         rating = s.get("safety_rating", "")
         rating_style = grade_styles.get(rating, "white")
-        updated = s.get("updated_at", "")[:10]
+        # ``dict.get(..., "")`` still returns ``None`` when the key is
+        # present with a null value (e.g. a row backfilled before the
+        # updated_at trigger fired); coerce with ``or ""`` so the slice
+        # cannot crash the entire table render.
+        updated = (s.get("updated_at") or "")[:10]
         table.add_row(
             s["org_slug"],
             s["skill_name"],
--- a/client/tests/test_cli/test_registry_cli.py
+++ b/client/tests/test_cli/test_registry_cli.py
@@ -727,6 +727,47 @@ class TestListCommand:
     @respx.mock
     @patch("dhub.cli.config.get_optional_token", return_value="test-token")
     @patch("dhub.cli.config.get_api_url", return_value="http://test:8000")
+    def test_list_handles_null_updated_at(
+        self,
+        _mock_url,
+        _mock_token,
+    ) -> None:
+        """Regression: rows with ``updated_at: null`` used to crash the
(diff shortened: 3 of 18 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; tracker disabled-count and ask-fallback analytics fixes catalogued in #451. Branch preserved.

PR #427, opened 2026-07-15
fix: correctness, security, and hardening pass across server + CLI

lfiaschi:
## What changed

A deep-review pass across the server, client, shared, and frontend packages surfaced a handful of concrete bugs and hardening opportunities. This PR ships the safest, highest-leverage subset with unit tests for every fix — 19 files touched, +534 / −114, all existing tests still green.

### Correctness bugs

- **`infra/database._row_to_skill_tracker`** was silently dropping `consecutive_permanent_failures`. Because the `SkillTracker` dataclass defaults the field to `0`, every tracker read from the DB reset the counter — the "disable after N permanent failures" backoff never triggered in practice.
- **`domain/evals.stream_eval_pipeline` / `run_streaming_eval`** decided pass/fail with `all(r["verdict"] == "pass" for r in case_results)`. `all([])` is `True`, so a skill shipping zero eval cases silently reported `completed`. Guarded with an explicit `bool(case_results)` check.
- **`infra/github_app_token.mint_installation_token`** set `exp = now + 600`, but GitHub caps `exp ≤ github_now + 600`. If the local clock is even seconds ahead of GitHub's, every mint returned 401 with no retry. Pull `exp` back to `now + 540` so the token has 60 s of clock-drift margin.

### Security

- **`api/rate_limit`** now honours `X-Forwarded-For`. Without it, all requests behind Modal's TLS terminator hashed to the proxy address, so one heavy caller starved everyone else on the same container.
- **`api/rate_limit`** lazy-init race: eight near-identical `_enforce_*_rate_limit` helpers were racing on first-touch. Two concurrent requests both failed the `hasattr` check, both instantiated a limiter, one `setattr` won — the losing bucket's accounting was silently discarded. Extracted `get_or_create_limiter` guarded by a module-level lock; every enforce dependency now goes through it. Same PR also replaces the `O(N-total-entries)` `sum()` on every admission with an `O(1)` accumulator that triggers `_purge_stale` every 100 admissions.
- **`logging._extract_username_from_jwt`** now returns `"unverified:<username>"` instead of the raw claim. The middleware doesn't verify the JWT signature, so anyone could forge a Bearer token with `{"username":"admin"}` and pollute logs with `admin`. The prefix keeps log-search results honest; verified identity is still available downstream via `get_current_user`.
- **`api/app.SecurityHeadersMiddleware`** now emits `Referrer-Policy: strict-origin-when-cross-origin` and a `Permissions-Policy` that denies camera / microphone / geolocation / payment / USB / MIDI / XR — features the SPA never uses. Prevents skill and eval URLs from leaking into third-party analytics via `Referer`.
- **`client/cli/config.save_config`** now writes atomically (temp-file + `os.replace`) and chmods the file to `0600` and the `~/.dhub` directory to `0700` on POSIX. Previous behaviour: a Ctrl-C mid-write left the config truncated; the loader treated that as corruption and asked the user to delete their config, losing their long-lived bearer token. Also plugs the "token file is world-readable" gap on multi-user systems.

### Performance

- **`api/auth_routes`** runs per-org GitHub membership checks concurrently via `asyncio.gather` instead of awaiting them sequentially. A user in the Nth allowed org previously paid `N × round-trip` latency on every login.

### Ergonomics

- **`infra/cache.TTLCache`** adds a `MISS` sentinel and a `default` parameter to `get()` so callers can distinguish "no entry" from "cached `None`" — required for callers that want to memoise a query that legitimately returns `None`. The default remains `None`, so no existing caller changes.

## Why

Standalone code-review pass — not tracking a specific issue. The findings all trace back to concrete failure modes documented in review notes and in the commit body.

## How to test

- `make test-server` — 947 tests pass (34 slow deselected).
- `make test-shared` — 98 tests pass.
- `make test-client` — new config tests pass; 5 pre-existing `docx_integration` failures on `main` are unrelated (fixture drift, verified by running against the base commit).
- `make lint` — clean.
- Targeted new tests:
  - `pytest server/tests/test_api/test_rate_limit.py -v` — concurrent lazy-init produces one instance; X-Forwarded-For splits per-real-client; purge accumulator sweeps stale keys.
  - `pytest server/tests/test_infra/test_cache.py::TestMissSentinel -v` — MISS disambiguates None-vs-absent.
  - `pytest server/tests/test_api/test_logging.py::TestExtractUsernameFromJwt -v` — forged `admin` token surfaces as `unverified:admin`.
  - `pytest server/tests/test_api/test_security_headers.py -v` — Referrer-Policy + Permissions-Policy present on both 200 and 404.
  - `pytest server/tests/test_domain/test_streaming_evals.py::TestReportEvent::test_zero_cases_marks_report_failed_not_completed -v`.
  - `pytest server/tests/test_infra/test_github_app_token.py -v` — `exp - iat == 10min`.
  - `pytest server/tests/test_infra/test_tracker_db.py::TestRowToSkillTracker::test_maps_consecutive_permanent_failures -v`.
  - `pytest client/tests/test_cli/test_config.py -v` — 0600 file perms, 0700 dir perms, atomic-write survival test.

## Checklist

- [x] Tests pass (`make test` for server + shared + client; pre-existing docx failures on main are unrelated)
- [x] No breaking API changes (public APIs unchanged; `TTLCache.get` default is still `None`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01EoPGA764JeRx89dW7PGec5)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -82,14 +83,37 @@ def load_config() -> CliConfig:
 def save_config(config: CliConfig) -> None:
     """Save CLI config to ~/.dhub/config.{env}.json.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ~/.dhub directory if it does not already exist. The
+    config file contains a long-lived bearer token, so:
+
+    - The directory is created with mode 0700 (owner-only) so a
+      brand-new install is not readable by other local users.
+    - The file is written atomically via a temp-then-``os.replace``
+      dance: Ctrl-C mid-write no longer leaves the config half-written
+      (which the loader treats as corrupted and asks the user to delete,
+      losing their token).
+    - The file itself is chmod'd to 0600 so an existing ~/.dhub with
+      loose permissions is tightened on next save. On Windows this is a
+      no-op because chmod semantics don't apply — POSIX users benefit,
+      Windows behaviour is unchanged.
     """
     CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+    # Tighten directory perms best-effort; ignore errors on Windows where
+    # POSIX mode semantics don't apply.
+    with contextlib.suppress(OSError, NotImplementedError):
+        CONFIG_DIR.chmod(0o700)
+
     path = config_file()
-    path.write_text(
-        json.dumps(asdict(config), indent=2) + "\n",
-        encoding="utf-8",
-    )
+    payload = json.dumps(asdict(config), indent=2) + "\n"
+
+    # Write to a sibling temp file first so a mid-write interrupt cannot
+    # leave the real config truncated. ``os.replace`` is atomic on POSIX
+    # and Windows (Python 3.3+).
+    tmp = path.with_name(path.name + ".tmp")
+    tmp.write_text(payload, encoding="utf-8")
+    with contextlib.suppress(OSError, NotImplementedError):
+        tmp.chmod(0o600)
+    os.replace(tmp, path)
 
 
 def get_api_url() -> str:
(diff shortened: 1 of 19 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its all([]) eval pass bug, GH-App token clock-drift margin, and SecurityHeadersMiddleware are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #428, opened 2026-07-16
refactor: leverage-focused fixes from architect+PE deep review

lfiaschi:
## What changed

Deep architect + principal-engineer review of the whole codebase (server, client, shared, frontend) plus a focused refactor landing the highest-leverage, low-risk fixes it turned up. 18 files, +496 / -201, all quality gates green. Nine distinct fixes in one commit — grouped so the review write-up below stays close to the diff.

## Why

Preventive audit — no single issue, several small ones that compound (SQL non-determinism, DOS gap behind the proxy, transient errors that don't actually retry soon, YAML fallback that corrupts real mappings, subprocess timeouts missing, unclear frontend retry loop, DRY on 9 near-identical helpers).

## How to test

```bash
make lint        # ruff check + format + frontend eslint
make typecheck   # mypy
make test        # client + server + shared + frontend
```

- Server: 938 pass, 34 slow deselected — new `test_rate_limit` XFF cases + `TestRateLimitDependency` factory tests + updated tracker tests.
- Client: 306 pass — new `test_app_installer.py` covers the first-token match, subprocess timeouts, and sibling-package rejection. (The 5 pre-existing `test_docx_integration` failures are local-skill drift on this container — origin fails the same way, unrelated to this diff.)
- Shared: 100 pass — new fallback-preserves-flow-mapping / flow-sequence tests.
- Frontend: 82 pass, `tsc -b` clean, `eslint` reports one pre-existing warning unrelated to this diff.

Manual smoke suggestions if reviewers want to poke:
- `curl -H "X-Forwarded-For: 1.2.3.4" http://.../health` behind a proxy still rate-limits per client once `rate_limit_trust_proxy=True` is set.
- Publishing a skill whose `SKILL.md` frontmatter has `description: Great: it works` and `metadata: {author: Alice}` — metadata now parses as a mapping instead of the literal string `"{author: Alice}"`.

## Checklist
- [x] Tests pass (`make test` — modulo the pre-existing docx integration drift noted above)
- [x] No breaking API changes (the new `keys` rate limiter and `rate_limit_trust_proxy` setting both default to safe values)
- [ ] Database migration included — not needed, no schema changes

---

## Deep-review summary

The review was run as two parallel Explore agents (server; client + shared + frontend), then synthesized. I picked the fixes below by leverage × safety — each is confined to one or two files and comes with tests. The bigger findings that need a design discussion (or a migration) are listed under **Deferred** at the bottom of this PR.

### Landed in this PR

| # | Category | Change | File · line |
|---|---|---|---|
| 1 | DRY / Architecture | Extract `rate_limit_dependency(name)` factory; replaces 9 near-identical `_enforce_*_rate_limit` helpers across `auth`, `search`, `registry`, `keys`. | `server/api/rate_limit.py:96`, `registry_routes.py:86` |
| 2 | Security / DOS | Rate limiter now bucket-keys off the left-most `X-Forwarded-For` when `rate_limit_trust_proxy` is on. Behind Modal every client was previously sharing one bucket because `request.client.host` == proxy IP. | `server/api/rate_limit.py:15` |
| 3 | Security / Throttling | `/v1/keys` (POST + GET + DELETE) now behind `_enforce_keys_rate_limit`. POST does Fernet encryption (CPU-bound) and a stolen token could enumerate at line rate. | `server/api/keys_routes.py:19` |
| 4 | SQL correctness | `search_skills_hybrid` (FTS + vector paths) and `fetch_similar_skills` gain a `skills.id` tiebreaker — LIMIT without a unique ORDER BY was letting pagination flip between requests when ranks tied. | `server/infra/database.py:2043,2055,2131` |
| 5 | SQL correctness | `find_active_eval_runs_for_user` gains an `id.desc()` tiebreaker for the same reason. | `server/infra/database.py:2731` |
| 6 | Bug / consistency | `fetch_skills_by_repo` was the only list-shape query missing `_exclude_removed_or_archived`, so `/repo` lookups leaked deleted-source / archived skills the rest of the UI already hides. | `server/infra/database.py:1971` |
| 7 | Bug / reliability | Transient tracker failures + circuit-breaker downgrades now route through `batch_defer_trackers` instead of `batch_set_tracker_errors`. Because `claim_due_trackers` bumps `next_check_at = now + poll_interval` on claim, the previous "transient — will retry" path had already been pushed a full poll interval into the future, making the log message misleading. Deferring clears `next_check_at` so the next tick picks them up. | `server/domain/tracker_service.py:312` |
| 8 | Bug / reliability | Every uv/pipx/pip subprocess in the CLI now has a bounded timeout (30s probes, 10min upgrade) and swallows `TimeoutExpired` / `FileNotFoundError` cleanly. Also introduces `_first_token_is_dhub_cli` — `line.startswith("dhub-cli")` also accepted `dhub-cli-extras` and returned its version. | `client/cli/app.py:99` |
| 9 | Correctness | `parse_frontmatter_yaml` fallback used to blanket-quote every colon-bearing value, so a real `metadata: {a: b}` YAML mapping got mangled into the string `"{a: b}"`. Now skips values that open a flow container / block scalar / anchor / alias (`{`, `[`, `|`, `>`, `&`, `*`, `!`, quotes). Markdown-link exception (`](`) still forces quoting because those don't close as a real flow sequence. | `shared/dhub_core/manifest.py:93` |
| 10 | UX clarity | `SkillDetailPage` zip-download retries are now driven by an explicit `zipAttempt` counter that `useEffect` observes. Old code relied on `setZipLoading(false)` implicitly re-firing the fetch effect — worked, but only by side-effect ordering. | `frontend/pages/SkillDetailPage.tsx:68` |

### Test additions

- `server/tests/test_api/test_rate_limit.py` — three new XFF cases (ignored by default, honored when trusted, empty XFF falls back to `client.host`) and two `rate_limit_dependency` tests (lazy init reuses the limiter; settings are read by name).
- `client/tests/test_cli/test_app_installer.py` — 15 tests covering `_first_token_is_dhub_cli`, `_run_installer` (missing binary → None, timeout → None), `_detect_installer` (uv, pipx, pip fallbacks + sibling rejection), `_query_version` (uv/pipx/pip parsing + skipped siblings).
- `shared/tests/test_manifest.py` — `test_fallback_preserves_flow_mapping`, `test_fallback_preserves_flow_sequence`.
- `server/tests/test_domain/test_tracker_service.py` — transient / circuit-breaker assertions moved to `batch_defer_trackers` to match the new behavior.
- `server/tests/test_api/conftest.py` — Settings mock advertises the new `keys_rate_limit` / `keys_rate_window` / `rate_limit_trust_proxy` fields so the `keys_routes` tests still exercise their real dependency.

### Full review findings — deferred for follow-up

Called out here so they're on record; happy to open individual issues if you want.

**Server**

- **`infra/database.py` is a 3,465-line god module** mixing users, orgs, skills, versions, api-keys, audit, eval reports, eval runs, trackers, batch-tracker, search-log, scan reports, embeddings, stats. Every route imports the whole thing. Suggested split: `infra/database/{users,orgs,skills,versions,trackers,audit,eval,scans}.py` + shared `_tables.py`. Left out of this PR because a 3k-line module split touches every route, has real merge-conflict cost, and is safer as its own PR with no functional changes.
- **`list_granted_skill_ids` called redundantly** on the same request (`resolve_version`, `resolve_latest_version`, `find_skill_by_slug`, `fetch_skills_by_repo`, `search_skills_hybrid` each re-fetch). Threading a `granted_skill_ids: list[UUID] | None` argument through the affected callers would collapse 4-5 lookups per request to 1. Left out because it touches many signatures at once.
- **`_check_zombie` runs an UPDATE inside GET handlers** (`registry_routes.py:1100`). Two concurrent polls to `/eval-runs/{id}` can both find a stale heartbeat and both run the UPDATE. A background reaper (or an explicit skip when the result is already stale) would be cleaner. Deferred — needs a design discussion about who owns the reaper.
- **Publish race: version row visible before S3 upload** (`publish_pipeline.py:752–768`). Between the `conn.commit()` that publishes the row and the `upload_skill_zip` call, a concurrent `resolve_skill` can hand a client a presigned URL that 404s. Fix is to upload first (staging key), then commit. Deferred because it reorders the pipeline and needs a careful test pass.
- **Long-lived JWT (default 8760h) with `github_orgs` never re-validated**. A removed org member keeps write access until the token expires. Introducing refresh + short-lived tokens + a `jwt_version` claim is a real change with a client-side migration path — clearly out of scope for this PR.
- **Rate-limiter memory-purge trigger uses `total % 100 == 0`**. Fine in practice but fragile; a heap-based sweep on a fixed interval would be cleaner. Very low priority.

**Client / shared**

- **CLI error output goes to stdout**, so `--output json` gets red-ANSI-decorated `"[red]Error: …"` written to stdout alongside the JSON payload. Every module defines its own `console = Console()` (stdout) — only `output.exit_error` writes to stderr. Fix: route all error prints through a shared helper that respects `is_json()` and writes to stderr. Left out because it touches ~15 files and every `console.print("[red]…"`) call site.
- **`_tail_eval_logs` polls forever** (`registry.py:1373`) — `while True: ... sleep(1.5)`. Bound total wall-clock + backoff. Small change, but I want to see how logs currently behave on the retry side before landing.
- **`_ensure_tracker` swallows non-typer exceptions silently** after a tracker GET/PATCH; a `500` from the server yields zero user-visible feedback.

**Frontend**

- **Hard-coded font sizes** in `GradeBadge.module.css:13,19,25`, `SkillDetailPage.module.css:381`, `AskModal.module.css:291` — violates the `--text-*` scale rule in `CLAUDE.md`. Small but noisy — group into its own PR.
- **Several components ship without `@media (max-width: 480px)` rules** despite appearing in card grids: `AnimatedTerminal`, `NeonCard`, `SkillCardStats`, `LoadingSpinner`, `GradeBadge`, `FileBrowser`. Also violates the mobile-first rule in `CLAUDE.md`.
- **`useApi.refetch`** has no staleness guard; a fast retry-click can commit stale results over new ones. Same `cancelled` flag pattern from the auto-fetch branch would fix it.

**Testing / observability**

- `execute_publish` has no direct unit tests — S3 upload failure rollback, `_try_store_scan_result` failure branches, `maybe_trigger_agent_assessment` when Modal is unavailable, and the auto-bump conflict path are all only exercised end-to-end.
- Eval-run zombie / heartbeat logic (`registry_routes.py:1100–1120`) is untested.
- Access-grant endpoints, similar-skills, `_log_ask_analytics` failure paths — no route-level coverage.

I'd suggest a follow-up PR for one of the two big items — either the `database.py` split or the JWT refresh flow. Everything else can be picked up as issues.

https://claude.ai/code/session_016xLDXxLB2JPD5a862yRaFQ

---
_Generated by [Claude Code](https://claude.ai/code/session_016xLDXxLB2JPD5a862yRaFQ)_

### DIFF
--- a/client/src/dhub/cli/app.py
+++ b/client/src/dhub/cli/app.py
@@ -96,26 +96,50 @@ from dhub.cli.doctor import doctor_command  # noqa: E402
 app.command("doctor")(doctor_command)
 
 
+_DHUB_CLI = "dhub-cli"
+# Any subprocess we spawn to talk to uv / pipx / pip could hang forever on a
+# broken network or stalled tool — the CLI should fall back rather than block.
+_INSTALLER_TIMEOUT = 30
+
+
+def _first_token_is_dhub_cli(line: str) -> bool:
+    """Return True when the first whitespace token of `line` is exactly `dhub-cli`.
+
+    `str.startswith("dhub-cli")` also matches sibling packages like
+    `dhub-cli-extras`, so match the first token exactly instead.
+    """
+    parts = line.split()
+    return bool(parts) and parts[0] == _DHUB_CLI
+
+
+def _run_installer(cmd: list[str]) -> subprocess.CompletedProcess | None:
+    """Run an installer command with a bounded timeout; return None on failure."""
+    try:
+        return subprocess.run(cmd, capture_output=True, text=True, timeout=_INSTALLER_TIMEOUT)
+    except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
+        return None
+
+
 def _detect_installer() -> str:
     """Detect how dhub-cli was installed: 'uv', 'pipx', or 'pip'."""
     uv_bin = shutil.which("uv")
     if uv_bin:
-        result = subprocess.run(
-            [uv_bin, "tool", "list"],
-            capture_output=True,
-            text=True,
-        )
-        if result.returncode == 0 and any(line.startswith("dhub-cli") for line in result.stdout.splitlines()):
+        result = _run_installer([uv_bin, "tool", "list"])
+        if (
+            result
+            and result.returncode == 0
+            and any(_first_token_is_dhub_cli(line) for line in result.stdout.splitlines())
+        ):
             return "uv"
 
     pipx_bin = shutil.which("pipx")
     if pipx_bin:
-        result = subprocess.run(
-            [pipx_bin, "list", "--short"],
-            capture_output=True,
-            text=True,
-        )
-        if result.returncode == 0 and any(line.startswith("dhub-cli") for line in result.stdout.splitlines()):
+        result = _run_installer([pipx_bin, "list", "--short"])
+        if (
+            result
+            and result.returncode == 0
+            and any(_first_token_is_dhub_cli(line) for line in result.stdout.splitlines())
+        ):
             return "pipx"
 
     return "pip"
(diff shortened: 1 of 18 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its /v1/keys rate limiting and fetch_skills_by_repo visibility fix are catalogued in #451. Branch preserved.

PR #430, opened 2026-07-18
perf: faster RateLimiter and DRY-up rate-limit boilerplate

lfiaschi:
## What changed

Three focused server-side hot-path cleanups, all bounded to the API layer and covered by tests. No migrations, no config changes, no CLI-facing endpoint shapes touched.

**1. `RateLimiter` internals** (`server/src/decision_hub/api/rate_limit.py`)

The per-IP sliding-window limiter previously did two expensive things on every request while holding the lock:

- Rebuilt the per-key timestamp list from scratch (`self._requests[key] = [t for t in timestamps if t > cutoff]`), allocating a new list per call.
- Walked the entire dict computing `sum(len(v) for v in ...)` to decide whether to trigger a "periodic" purge, which was actually driven by `total % 100 == 0` — a workload-dependent trigger that fires arbitrarily often or not at all under real traffic.

Rewritten to use a `deque` per key with `popleft()` while the head is stale (amortised O(1), no allocation), and a plain integer counter for the purge cadence (`_PURGE_EVERY_N_REQUESTS = 500`). The public API (`__init__`, `__call__`) is unchanged so all eight call sites need no updates. Purge now fires on a truly periodic schedule regardless of workload distribution.

**2. Consolidate `_enforce_*_rate_limit` boilerplate**

The API layer had nine copies of the same 8-line lazy-init pattern (seven in `registry_routes.py` plus one each in `search_routes.py` and `auth_routes.py`). Factored into `make_rate_limit_dependency(name)` on the rate-limit module — each endpoint's guard collapses to a single line:

```python
_enforce_publish_rate_limit = make_rate_limit_dependency("publish")
```

Adding a new limited endpoint no longer requires copy-pasting a whole block. Removes ~90 lines net; `Request` / `Settings` imports in the two smaller files are now unused and dropped.

**3. Skip redundant DB round-trip in `get_eval_run`**

`_check_zombie` used to write the zombie status to the DB and let the route re-query the row unconditionally, wasting a query on the happy path (no zombie detected). Since `EvalRun` is a frozen dataclass, `_check_zombie` now returns a `dataclasses.replace(...)` copy reflecting the fields it wrote and the route uses that directly — **one DB round-trip saved per `GET /v1/eval-runs/{id}` call**. The CLI polls this endpoint every ~1 s during an eval, so it's a real hot path. The `logs` endpoint's identical pattern is updated the same way.

## Why

Standalone review pass — no linked issue. Findings from a focused code-review sweep over the API hot paths.

## How to test

- `make test-server` (or `pytest server/tests -m "not slow"`) — 940 tests pass locally.
- `uvx ruff@0.15.3 check server/src server/tests` — clean.
- `uvx mypy server/src` — clean.

New / updated coverage:

- `server/tests/test_api/test_rate_limit.py` — deque storage, in-place reuse, deterministic purge cadence, and the `make_rate_limit_dependency` factory (lazy init, sharing, enforcement, `__name__` propagation).
- `server/tests/test_api/test_eval_logs.py::test_zombie_detection_marks_stale_run_as_failed` — now asserts that only **one** call to `find_eval_run` happens and that the response's `error_message` / `completed_at` reflect the in-memory update.

## Checklist

- [x] Tests pass (`make test-server`, 940 passed)
- [x] No breaking API changes — response schemas, CLI-visible fields, and route signatures are unchanged
- [x] Database migration included (if schema changed) — no schema changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01CVMZ7aY2HQqzcJGjXqwXY8)_


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Strict subset of the rate-limiter work in retained #447. Branch preserved.

PR #431, opened 2026-07-19
Fix 10 concrete bugs across CLI, storage, tracker, and rate limiter

lfiaschi:
## What changed

Ten targeted bug fixes across the CLI, server storage/API/tracker layers, and the rate limiter — each backed by a new or updated regression test that fails on `main` and passes with this diff.

## Why

An architect-level review of the codebase (parallel deep-dive over `database.py`, `registry_routes.py`, CLI `registry.py`, `tracker_service.py`, `gauntlet.py`, `evals.py`, `search_routes.py`, `gemini.py`, `modal_client.py`, `storage.py`, and the frontend hooks) surfaced ~45 concrete findings. This PR takes the highest-leverage bug-fix subset that is coherent enough for one review pass; each fix has a concrete failing scenario, not a vague concern.

Larger structural cleanups (splitting the 1912-line CLI `registry.py`, deduplicating the 7 copy-paste rate-limit factories in `registry_routes.py`, the 3 near-copy LLM judgment functions in `gemini.py`, adding pagination-aware tests for `list_eval_runs` zombie detection, etc.) are deliberately out of scope here so the diff stays reviewable — they can land as follow-ups.

### CLI (`client/src/dhub/cli/registry.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| `_render_skills_table` coalesces `None → ""` before slicing `updated_at` | Server sends `"updated_at": null` (not absent) → `.get("updated_at", "")[:10]` returns `None[:10]` → `TypeError` → every `dhub list` crashes until backfill |
| `_try_resolve_run_id` no longer falls back to the user's global recent runs | `dhub logs alice/skill-a --follow`; skill-a has no eval report → CLI silently tails the user's newest run — potentially `bob/unrelated-skill` — with no indication |
| `_install_single_skill` uses `_download_capped` (200 MB streaming cap + Content-Length preflight) | Misconfigured registry / hostile mirror serving a multi-GB response OOMs the client silently before `zipfile` can refuse it |

### Server storage (`server/src/decision_hub/infra/storage.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| `list_eval_log_chunks` now uses the S3 paginator | An eval run with more than 1000 chunks (long or high-verbosity) silently drops every chunk past the first page; the resumable `after_seq` tail never sees the missing events |
| `delete_eval_logs` paginates the list and batches `delete_objects` in groups of 1000 | Same 1000-key cap plus the `delete_objects` per-call max → orphan objects left in S3 forever on cleanup |

### Server API (`server/src/decision_hub/api/registry_routes.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| `_check_zombie` falls back to `created_at` when `heartbeat_at IS NULL` | Worker dies before it ever writes a heartbeat → `_check_zombie` returned early → run stays `status="running"` in the UI forever |

### Server DB (`server/src/decision_hub/infra/database.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| `_row_to_skill_tracker` now maps `consecutive_permanent_failures` | Every `SkillTracker` read from the DB defaulted the field to `0` regardless of the actual column value; any admin/UI code reading `tracker.consecutive_permanent_failures` would be silently wrong |

### Server tracker (`server/src/decision_hub/domain/tracker_service.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| `tracker_to_dict` uses `isinstance(value, UUID)` instead of `hasattr(value, "hex")` | Any future byte-valued field on `SkillTracker` (e.g. a checksum column) would be silently corrupted to `"b'\x00...'"` on the Modal wire because `bytes` also has `.hex()` |
| `process_tracker` surfaces partial failures with a `partial: ...` prefix in `last_error` | 1 of 6 skills publishes, 5 fail; previously `all_failed=False` cleared `last_error`, the SHA advanced, and the 5 broken skills never retried and never showed in tracker health |
| `_detect_removed_skills` bails on any per-dir parse failure (not only when all fail) | One transient YAML syntax error in a single `SKILL.md` on a commit leaves the affected skill out of `discovered_names`; its DB row gets marked `source_repo_removed=true` and won't be resurrected by parseable follow-up commits |

### Server rate limit (`server/src/decision_hub/api/rate_limit.py`)

| Fix | Concrete failing scenario |
|-----|--------------------------|
| Purge trigger replaced with a request-count counter (`_PURGE_EVERY = 128`) | Previous `sum(len(v) for v in ...) % 100 == 0` was O(N) on the hot path (defeating the purpose) *and* skipped whenever the total happened to miss a multiple of 100, so stale IPs could accumulate unbounded |

## How to test

```bash
# Server (all 941 pass)
uv run --package decision-hub-server --extra dev pytest server/tests/ -m "not slow"

# Client (298 pass; the 5 failures in test_docx_integration.py are
# pre-existing on main — verified via `git stash`)
uv run --package dhub-cli --extra dev pytest client/tests/ \
    --ignore=client/tests/test_core/test_docx_integration.py

# Shared (98 pass)
uv run --package dhub-core --extra dev pytest shared/tests/

# Lint + format
uvx ruff@0.15.3 check .
uvx ruff@0.15.3 format --check .
```

The new regression tests live in:
- `client/tests/test_cli/test_registry_regressions.py` (new file — 3 classes covering the CLI fixes)
- `server/tests/test_api/test_eval_logs.py` — `test_zombie_detects_null_heartbeat_via_created_at`, `test_fresh_run_with_null_heartbeat_not_marked_zombie`
- `server/tests/test_api/test_rate_limit.py` — `test_periodic_purge_removes_stale_ip_entries`
- `server/tests/test_infra/test_storage_eval_logs.py` — `test_paginates_beyond_first_page`, `test_batches_delete_calls_in_groups_of_1000`
- `server/tests/test_infra/test_tracker_db.py` — `test_maps_consecutive_permanent_failures`
- `server/tests/test_domain/test_tracker_service.py` — `test_partial_parse_failure_skips_removal`, `test_uuid_conversion_is_isinstance_not_hasattr`; updated `test_partial_success_advances_sha` to assert the new `partial:` surface

Each test was verified to fail against the pre-fix code (checked by isolating the fix, running the test, and observing the failure mode described in the table above).

## Checklist
- [x] Tests pass (server / client / shared)
- [x] No breaking API changes (only user-visible change is that `dhub list` no longer crashes on `null updated_at`, and `dhub logs org/skill` returns a clean "not found" instead of silently tailing another skill's run)
- [x] No schema change → no migration needed
- [x] Lint + format clean (`ruff check .`, `ruff format --check .`)

https://claude.ai/code/session_01UhmKY33YeX1aRLHcUknEaC

---
_Generated by [Claude Code](https://claude.ai/code/session_01UhmKY33YeX1aRLHcUknEaC)_

### DIFF
--- a/client/src/dhub/cli/registry.py
+++ b/client/src/dhub/cli/registry.py
@@ -15,6 +15,41 @@ from rich.table import Table
 
 console = Console()
 
+# Hard cap on downloaded skill zip size. A misconfigured registry (or a hostile
+# mirror serving an oversized redirect) could otherwise stream gigabytes into
+# `resp.content` before `zipfile` refuses to parse it, OOMing the client.
+# 200 MB is well above any realistic published skill (largest today are <10 MB)
+# but small enough that clients on constrained hardware fail cleanly.
+_MAX_SKILL_ZIP_BYTES = 200 * 1024 * 1024
+
+
+def _download_capped(client: httpx.Client, url: str, max_bytes: int) -> bytes:
+    """GET `url` and return the body, aborting if it exceeds `max_bytes`.
+
+    Streams so we never allocate more than one chunk beyond the running
+    total, and checks Content-Length up front when present so we can fail
+    before spending bandwidth on an obviously oversized response.
+    """
+    with client.stream("GET", url) as resp:
+        resp.raise_for_status()
+        content_length = resp.headers.get("content-length")
+        if content_length is not None:
+            try:
+                declared = int(content_length)
+            except ValueError:
+                # Bogus header — ignore and let the streaming guard below
+                # (authoritative) decide.
+                declared = None
+            if declared is not None and declared > max_bytes:
+                raise ValueError(f"Response too large: server reported {declared} bytes (limit {max_bytes}).")
+
+        buf = bytearray()
+        for chunk in resp.iter_bytes():
+            buf.extend(chunk)
+            if len(buf) > max_bytes:
+                raise ValueError(f"Response too large: exceeded {max_bytes} bytes while downloading.")
+        return bytes(buf)
+
 
 def _publish_skill_directory(
     path: Path,
@@ -656,7 +691,9 @@ def _render_skills_table(skills: list[dict], title: str = "Published Skills") ->
     for s in skills:
         rating = s.get("safety_rating", "")
         rating_style = grade_styles.get(rating, "white")
-        updated = s.get("updated_at", "")[:10]
+        # `updated_at` may be explicitly `null` in the JSON payload (not just absent),
+        # in which case `.get("updated_at", "")` still returns None. Coalesce to "".
+        updated = (s.get("updated_at") or "")[:10]
         table.add_row(
             s["org_slug"],
             s["skill_name"],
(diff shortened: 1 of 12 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its capped-download and run-id-fallback fixes are catalogued in #451. Branch preserved.

PR #432, opened 2026-07-20
Security + correctness fixes surfaced by full-repo review

lfiaschi:
## What changed

Twelve high-signal fixes surfaced by a four-pronged code review (server domain, server API/infra, client CLI, frontend + tests). Focus is on security holes, correctness bugs, and one flaky-test category — no behavior changes on healthy paths.

## Why

Standalone review pass (no linked issue) — findings are enumerated below with the concrete failing scenarios that motivated each fix.

### Security

- **`client/core/git_repo.py` — git argument injection.** `git clone` and `git checkout` now include a `--` separator and reject `-`-prefixed URLs/refs early. A repo URL like `--upload-pack=foo.git` slipped past `looks_like_git_url` and let a malicious repo execute arbitrary commands during `dhub publish`.
- **`client/cli/config.py` — world-readable auth token.** `save_config` now writes `~/.dhub/config.*.json` with `0o600` via `os.open` (mode at creation time, not after) and the parent dir with `0o700`. On shared machines the file inherited the default umask (often 644), exposing the user's GitHub OAuth token to any other UID.
- **`server/api/rate_limit.py` — shared rate-limit bucket behind proxy.** Read `X-Forwarded-For` (leftmost hop) with `X-Real-IP` fallback so per-IP buckets bind to the real client, not the Modal proxy IP that every client shares. Without this, one hostile client could lock every public endpoint for everyone else on the same container.

### Correctness

- **`server/domain/publish_pipeline.py` — aborted-transaction leak.** `insert_version` raising `IntegrityError` now triggers `conn.rollback()` before we raise `VersionConflictError`. Postgres aborts the whole transaction on IntegrityError; any subsequent statement on the same connection would fail with "current transaction is aborted".
- **`server/domain/evals.py` — permanent chunk-seq gap on S3 failure.** `_flush_buffer` no longer advances `chunk_seq` before the S3 upload. Extracted into `_upload_next_chunk`, which returns the new seq only on success. Previously a transient S3 error left a permanent gap so the frontend saw a "missing" chunk that had in fact never been written.
- **`client/cli/registry.py` — install/update never removed stale files.** `_install_single_skill` now extracts to a sibling staging directory and atomically swaps it into place, so files the upstream skill has since deleted are actually removed. `zf.extractall` was merging new-and-old side by side forever.
- **`server/domain/gauntlet.py` — YAML-list `allowed_tools` silently dropped.** `allowed_tools: [Bash, WebFetch]` (perfectly valid YAML) is now coerced to a space-joined string instead of `None`. Previously such manifests looked like "no declared tools" and got flagged as inconsistent with the skill's elevated permissions.
- **`server/api/registry_routes.py` + `server/infra/database.py` — Python-side user filter on eval-runs.** Pushed the `user_id` predicate for `find_eval_runs_for_version` into Postgres. Previously the route fetched every row for the version and filtered in Python, leaking memory on hot versions and briefly holding other users' rows in server memory.

### Frontend

- **`pages/SkillDetailPage.tsx` —** Download button reuses the zip bytes already loaded for the Overview/Files tabs instead of re-fetching `/download`. Cleans up the `handleCopyInstall` `setTimeout` via a ref so we don't `setState` on an unmounted tree.
- **`pages/NotFoundPage.tsx`, `OrgDetailPage.tsx`, `components/FileBrowser.tsx` —** replace hardcoded rem font sizes with the project's `--text-*` custom properties (mobile-first typography scale).

### Testing

- **`server/infra/cache.py` + `tests/test_infra/test_cache.py` —** `TTLCache` now accepts an injectable monotonic clock so TTL-expiry tests advance a stub instead of sleeping. Removes real `time.sleep(0.01–0.02)` calls that flaked on loaded CI runners.
- Adds regression tests: `test_git_repo.py::TestValidateRef`, `test_config.py::test_config_file_is_not_world_readable`, `test_rate_limit.py::test_uses_x_forwarded_for_leftmost` (+ `test_x_real_ip_fallback`), `test_evals.py::TestUploadNextChunk`, and updates the gauntlet `allowed_tools` test to assert list-coercion (previously asserted the buggy drop-to-`None`).

## How to test

- `make test` — 939 server + 296 client + 98 shared + 82 frontend all green locally on this branch (excluding 5 pre-existing `test_docx_integration.py` failures unrelated to this diff — verified against `origin/main` HEAD).
- Manually: `dhub install org/skill` after publishing a version with a removed file should no longer leave that file around locally.
- Manually: hit `/v1/eval-runs?version_id=…` as one user for a version another user has runs on — response is empty, DB call is filtered on `user_id`.

## Checklist

- [x] Tests pass (`make test`) — see note above about pre-existing docx failures.
- [x] No breaking API changes.
- [x] Database migration included (if schema changed) — N/A, no schema changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TS8fb2jFt1tp8vVjAha9AL)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -80,16 +81,30 @@ def load_config() -> CliConfig:
 
 
 def save_config(config: CliConfig) -> None:
-    """Save CLI config to ~/.dhub/config.{env}.json.
+    """Save CLI config to ~/.dhub/config.{env}.json with 0o600 permissions.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ~/.dhub directory (mode 0o700) if it does not already
+    exist. The config file may contain the user's GitHub OAuth token,
+    so it must never be world- or group-readable on a shared machine.
     """
-    CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+    CONFIG_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
+    # Tighten permissions on ~/.dhub even if it already existed with a
+    # looser mode (previous CLI versions relied on the default umask).
+    # Best-effort: chmod may fail on some filesystems (e.g. certain network
+    # mounts). Don't block config writes over it.
+    with contextlib.suppress(OSError):
+        os.chmod(CONFIG_DIR, 0o700)
+
     path = config_file()
-    path.write_text(
-        json.dumps(asdict(config), indent=2) + "\n",
-        encoding="utf-8",
-    )
+    payload = json.dumps(asdict(config), indent=2) + "\n"
+    # Open with O_CREAT | O_TRUNC and mode 0o600 so a freshly-created
+    # file is never briefly world-readable between write and chmod.
+    fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+    with os.fdopen(fd, "w", encoding="utf-8") as fh:
+        fh.write(payload)
+    # Ensure existing files that predate this behavior get tightened too.
+    with contextlib.suppress(OSError):
+        os.chmod(path, 0o600)
 
 
 def get_api_url() -> str:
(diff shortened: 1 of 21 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its IntegrityError rollback, chunk_seq-on-success, and atomic install swap are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #433, opened 2026-07-21
fix: harden CLI, server, and frontend against a batch of latent defects

lfiaschi:
## What changed

A batch of low-risk, high-signal defect fixes across the CLI (`client/`, `shared/`), the server package, and the React frontend. Findings surfaced by a targeted architectural review; higher-effort items (see **Deferred** below) are called out for a follow-up PR rather than landed here.

**Client / shared**
- `git_repo.clone_repo` inserts `--` between `git clone` / `git checkout` and any user-controlled arg, so a `repo_url` that starts with `-` cannot hijack git as an option (`--upload-pack=<binary>` → RCE). Both subprocess calls also pick up an explicit `timeout` and `TimeoutExpired` is translated to `RuntimeError`, so a hung remote or SSH key prompt can't stall the CLI forever.
- Missing f-string in `registry.py:158` — the placeholder in `"Detached. Resume with: dhub logs {eval_run_id} --follow"` was being printed literally, so the resume command couldn't be copy-pasted.
- `shared.dhub_core.manifest.parse_skill_md` now reads `SKILL.md` as UTF-8 explicitly. On Windows the default cp1252 encoding raised on any skill with non-ASCII in its description.
- `parse_frontmatter_yaml` surfaces its fallback-parse failure as a `ValueError` so callers that already catch `ValueError` see a consistent error type (a raw `yaml.YAMLError` previously leaked past the caller's handler). `gauntlet.check_manifest_schema` updated to accept both.

**Server**
- `infra.database.batch_update_github_stars` / `batch_update_github_repo_metadata` now escape LIKE wildcards. A repo containing `_` or `%` previously stamped stars / forks / license / archived state onto unrelated skills (e.g. `my_repo` would match `myXrepo/*`).
- Every `ORDER BY … LIMIT` that previously ordered on a non-unique column (FTS rank, vector cosine distance, similar skills, active eval runs, tracker metrics) gets an `id ASC` tiebreaker so pagination and "similar skills" no longer drift across calls when values tie.
- `infra.storage.list_eval_log_chunks` and `delete_eval_logs` use a boto3 paginator instead of a single `list_objects_v2` call. Long-running eval logs (>1000 chunks) previously lost their tail — including the "run complete" chunk — and cleanup silently orphaned 2000+ S3 objects. `delete_eval_logs` also inspects the response's `Errors[]` and raises so partial failures stop being reported as success.
- `registry_routes.get_eval_run_logs` no longer 500s forever when a single JSONL line is malformed (e.g. from a worker crash mid-write) — the bad line is logged and skipped so the UI keeps making progress.
- Both `/resolve` and `/download` now `increment_skill_downloads` **after** the S3 operation succeeds, so failed presigns / downloads don't inflate the ranking counter.

**Frontend**
- `api/client.ts` wraps every path segment in `encodeURIComponent`. Org slugs or skill names with `%`, `/`, or spaces previously produced malformed URLs that 404'd or hit the wrong route.
- `GradeBadge` tolerates `null` / `undefined` / `""` and renders the pending indicator instead of throwing on `grade.trim()`. The DB column is nullable, so any skill still awaiting a scan was crashing the row.

**Tests**
- `clone_repo`: assert `--` is present in both the default and SHA clone command lines, that `timeout` is passed, and that `TimeoutExpired` becomes a `RuntimeError`.
- `list_eval_log_chunks`: assert pagination across multiple pages returns all 1500 chunks.
- `delete_eval_logs`: assert batching of >1000 keys and that a non-empty `Errors[]` raises.
- `GradeBadge`: `null` / `undefined` / `""` render pending instead of crashing.
- `parse_frontmatter_yaml`: assert the `ValueError` contract on unparseable input.
- Existing `check_manifest_schema` test still passes because the caller now catches both exception types.

## Why

Prevent security, correctness, and UX regressions surfaced by a codebase review:

- The unquoted `git clone` argument is an RCE vector on any host running `dhub publish <url>`.
- LIKE-wildcard bleed silently corrupts GitHub-metadata on unrelated skills — the sort of thing that is very hard to notice after the fact.
- The `/logs` endpoint 500-forever on one bad JSONL line was a real availability bug for any customer whose worker crashed at the wrong moment.
- Path segments not being URL-encoded is a `useParams`-of-doom waiting to happen.
- The nullable `safety_rating` grade crashing a react row is a live "unknown skill on the detail page" outage vector.

## How to test

```
uv run --package dhub-cli --extra dev pytest client/tests/ -q
uv run --package dhub-core --extra dev pytest shared/tests/ -q
cd server && DHUB_ENV=dev uv run --package decision-hub-server --extra dev pytest tests/ -q -m "not slow"
cd frontend && npx vitest run
uvx ruff@0.15.3 check client/ server/ shared/
uvx ruff@0.15.3 format --check client/ server/ shared/
uvx mypy client/src/ server/src/ shared/src/
```

Locally observed on this branch:
- 315 client tests pass. Five failures in `test_docx_integration` (file-count / description assertions) reproduce on `main` at the base commit — pre-existing, unrelated, left alone here.
- 98 shared tests pass.
- 936 server tests pass (`-m "not slow"`).
- 86 frontend vitest tests pass.
- Ruff, ruff-format, and mypy all clean.

## Deferred (call-outs for a follow-up PR)

These surfaced in the same review but are meaningful enough to want their own PR + design conversation rather than sneaking in with a low-risk batch:

- **JWT-baked `github_orgs` is not re-verified** in `create_organisation` (`api/org_routes.py`) — a user removed from a GitHub org can still claim the DB namespace until any real member logs in and the JWT is refreshed. Fix: call `check_org_membership` inside the handler.
- **Lost-update race on `skills.latest_semver` / `latest_version_id`** in `insert_version` / `delete_version` (`infra/database.py`) — concurrent publishes can stamp the wrong "latest". Fix: fold into a single `UPDATE ... SET latest_semver = (SELECT ... LIMIT 1)`.
- **Rate limiter keys off `request.client.host`** (`api/rate_limit.py`) which behind Modal's ingress is the proxy IP; effectively no per-IP limit. Fix: parse `X-Forwarded-For` (with a proxy allowlist).
- **Rate limiter lazy-init race** (`api/*_routes.py`) — `hasattr` check-then-set with no lock in sync deps that run in the threadpool; two cold-start requests each create a fresh `RateLimiter` and lose one's counters. Fix: init eagerly at app startup or guard with a lock.
- **DB pool exhaustion via GH HTTP inside `engine.begin()`** in `tracker_routes.py:187` — `check_repo_accessible` runs inside the transaction; concurrent POSTs during a GH slowdown exhaust the pool. Fix: perform the HTTP before opening the tx.

## Checklist

- [x] No breaking API changes
- [x] Database migration included (if schema changed) — n/a

---
_Generated by [Claude Code](https://claude.ai/code/session_016EYSftjC5byFADmZUM8za2)_

### DIFF
--- a/client/src/dhub/cli/registry.py
+++ b/client/src/dhub/cli/registry.py
@@ -155,7 +155,7 @@ def _publish_skill_directory(
 
             _tail_eval_logs(api_url, _bh(token), eval_run_id)
         except KeyboardInterrupt:
-            console.print("\n[dim]Detached. Resume with: dhub logs {eval_run_id} --follow[/]")
+            console.print(f"\n[dim]Detached. Resume with: dhub logs {eval_run_id} --follow[/]")
     elif eval_report_status == "pending":
         console.print("[dim]Agent assessment running in background...[/]")
 
--- a/client/src/dhub/core/git_repo.py
+++ b/client/src/dhub/core/git_repo.py
@@ -66,32 +66,40 @@ def clone_repo(repo_url: str, ref: str | None = None) -> Path:
     tmp_dir = Path(tempfile.mkdtemp(prefix="dhub-repo-"))
     repo_path = tmp_dir / "repo"
 
-    if ref and _looks_like_sha(ref):
-        # Commit SHAs don't work with --depth 1 --branch; do a full
-        # clone then checkout the specific commit.
-        cmd = ["git", "clone", repo_url, str(repo_path)]
-        result = subprocess.run(cmd, capture_output=True, text=True)
-        if result.returncode != 0:
-            shutil.rmtree(tmp_dir, ignore_errors=True)
-            raise RuntimeError(f"git clone failed (exit {result.returncode}):\n{result.stderr.strip()}")
-        checkout = subprocess.run(
-            ["git", "checkout", ref],
-            cwd=str(repo_path),
-            capture_output=True,
-            text=True,
-        )
-        if checkout.returncode != 0:
-            shutil.rmtree(tmp_dir, ignore_errors=True)
-            raise RuntimeError(f"git checkout {ref} failed:\n{checkout.stderr.strip()}")
-    else:
-        cmd = ["git", "clone", "--depth", "1"]
-        if ref:
-            cmd += ["--branch", ref]
-        cmd += [repo_url, str(repo_path)]
-        result = subprocess.run(cmd, capture_output=True, text=True)
-        if result.returncode != 0:
-            shutil.rmtree(tmp_dir, ignore_errors=True)
-            raise RuntimeError(f"git clone failed (exit {result.returncode}):\n{result.stderr.strip()}")
+    # The `--` separator prevents any repo_url that starts with `-` from being
+    # parsed as a git option (which would allow argument injection like
+    # --upload-pack=<attacker binary>). Same reasoning for `ref` on checkout.
+    try:
+        if ref and _looks_like_sha(ref):
+            # Commit SHAs don't work with --depth 1 --branch; do a full
+            # clone then checkout the specific commit.
+            cmd = ["git", "clone", "--", repo_url, str(repo_path)]
+            result = subprocess.run(cmd, capture_output=True, text=True, timeout=120)
+            if result.returncode != 0:
+                shutil.rmtree(tmp_dir, ignore_errors=True)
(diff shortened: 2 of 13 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its encodeURIComponent path guard and GradeBadge null-safety are catalogued in #451. Branch preserved.

PR #434, opened 2026-07-22
fix: correctness + observability fixes from architectural review

lfiaschi:
## What changed

A first pragmatic pass from a full-repo audit (server + client + shared + frontend). Concrete bug fixes with regression tests, a small dedup, and one observability fix. Every change has a documented failing scenario; nothing here is stylistic.

**Server correctness**
- `database.batch_update_github_stars` / `batch_update_github_repo_metadata` — escape `LIKE` wildcards. A repo URL containing `_` (very common — e.g. `pymc_labs`) was treated as a single-char wildcard and could clobber unrelated repos' star/fork/watcher counts.
- `github.check_org_membership` — raise `HTTPStatusError` on 429/5xx instead of silently returning `False`. Silent `False` meant a GitHub rate-limit or outage locked legitimate members out of login. `auth_routes` now surfaces a `502` (retryable) for transient errors and keeps `403` only for genuine non-membership.
- `storage.list_eval_log_chunks` / `delete_eval_logs` — paginate `list_objects_v2` via boto3's paginator, and batch `delete_objects` at the 1000-key limit. Previously a long eval flushing more than 1000 chunks silently lost its tail.
- `publish.extract_for_evaluation` — decode scannable source files with `errors="replace"` so a Latin-1 / CP1252 file in a third-party repo can't abort the whole publish with `UnicodeDecodeError`. `SKILL.md` (the manifest) stays strict.
- `registry_routes.get_eval_run` — use `_check_zombie` return status directly instead of re-reading the row. Saves one SELECT and avoids a rare `NoneType` crash if the row is deleted between the two reads.
- `registry_routes.list_eval_runs` + `database.find_eval_runs_for_version` — push `user_id` filter into SQL instead of pulling every run for a version across all users and filtering in Python.
- `backfill_embeddings` — add `ORDER BY skills.id` so the `LIMIT` scan is deterministic (per CLAUDE.md invariant).

**Observability**
- `RequestLoggingMiddleware` — on unhandled exceptions attach the traceback via `logger.opt(exception=...)` so the request summary line carries the stack instead of a bare "500" with no correlation.

**Code quality**
- Extract `parse_uuid` into `api/deps.py`; drop the two duplicated `_parse_uuid` copies (`registry_routes.py` and `tracker_routes.py`) and their now-unused `from uuid import UUID` imports.

**Frontend**
- `AskModal` — reset `messages` / `query` / `error` when the modal closes so a subsequent open starts fresh instead of resurfacing the prior session.
- `GradeBadge.module.css` / `NotFoundPage.tsx` — use the `--text-*` CSS custom-property scale instead of hardcoded `rem` sizes.

## Why

Findings surfaced by a scheduled review pass across the codebase. Each fix corresponds to a bug with a specific reproduction case (e.g. GitHub rate-limit → auth denial; underscore in repo URL → wrong repo's stars overwritten; long eval → truncated log tail); each is small enough to review in isolation and covered by a regression test.

## How to test

```
make test-server    # 950 passed
make test-shared    # 98 passed
make test-frontend  # 82 passed
make lint           # ruff + frontend clean (1 pre-existing warning)
make typecheck      # mypy clean
```

New / updated tests:

- `test_batch_update_github` — LIKE-escape regression for `_` / `%` in repo URLs.
- `test_github.TestCheckOrgMembership` — 204/404/302/403 (False) vs 429/500/502 (raises) matrix.
- `test_storage_eval_logs` — pagination beyond 1000 keys; delete batching at the 1000-key API cap.
- `test_publish.test_survives_non_utf8_scannable_file` — Latin-1 regression.
- `test_publish.test_skill_md_still_strict_on_bad_utf8` — guards against loosening `SKILL.md` decoding by accident.
- `test_eval_logs.test_list_by_version_filters_to_current_user` — updated to assert the `user_id` kwarg is passed to the DB helper (filter is now in SQL, not Python).

Manual: unchanged surface for CLI + web UI. No schema changes, no migrations.

## Checklist
- [x] Tests pass (`make test-server`, `make test-shared`, `make test-frontend`)
- [x] No breaking API changes
- [x] Database migration included (if schema changed) — not needed; no schema changes

## Not in this PR (called out in the review)

Larger-blast-radius findings that need a design conversation before landing:

- **Rate limiter reads Modal proxy IP, not client IP**, and lives per-container — per-IP limits are effectively decorative on Modal, and auth brute-force isn't actually rate-limited across replicas. Needs a trust decision on which forwarded-header to honour and a shared backend (Redis / Modal Dict).
- **Batch tracker `UPDATE` uses N per-row round-trips per tick**. At 10k+ trackers this is a real cost; rewrite as one CTE-based UPDATE with `VALUES (...)` unnest.
- **`get_eval_run_logs` reads every S3 chunk on every poll** (comment acknowledges `cursor` is an event-seq, not a chunk-seq). Fix needs min/max-seq metadata baked into the S3 key so early chunks can be skipped.
- **Publish pipeline has no direct tests** — coverage is via 12–15 deep `@patch` stacks in `test_registry_routes.py`, which are refactor-hostile.
- **9 near-identical `_enforce_*_rate_limit` stubs** across route files should collapse into one factory in `api/rate_limit.py`.
- **`crawler.processing._finalize_skill` uploads to S3 before `insert_version`** — on `IntegrityError` from a race the blob is orphaned. `execute_publish` already gets the ordering right; mirror it.

https://claude.ai/code/session_01Rgm9rE4S3GW9xDZwAJ56wR

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rgm9rE4S3GW9xDZwAJ56wR)_

### DIFF
--- a/frontend/src/components/AskModal.tsx
+++ b/frontend/src/components/AskModal.tsx
@@ -46,6 +46,16 @@ export default function AskModal({ isOpen, onClose }: AskModalProps) {
     if (isOpen) refreshRecentlyViewed();
   }, [isOpen, refreshRecentlyViewed]);
 
+  // Reset conversation state when the modal closes so a subsequent open
+  // starts fresh instead of resurfacing the previous session's messages/error.
+  useEffect(() => {
+    if (!isOpen) {
+      setMessages([]);
+      setQuery("");
+      setError(null);
+    }
+  }, [isOpen]);
+
   // Scroll to bottom when new messages arrive
   useEffect(() => {
     messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
--- a/frontend/src/components/GradeBadge.module.css
+++ b/frontend/src/components/GradeBadge.module.css
@@ -10,19 +10,19 @@
 }
 
 .sm {
-  font-size: 0.65rem;
+  font-size: var(--text-xxs);
   padding: 2px 8px;
   min-width: 28px;
 }
 
 .md {
-  font-size: 0.8rem;
+  font-size: var(--text-xs);
   padding: 4px 12px;
   min-width: 36px;
 }
 
 .lg {
-  font-size: 1.4rem;
+  font-size: var(--text-heading);
   padding: 8px 20px;
   min-width: 56px;
 }
--- a/frontend/src/pages/NotFoundPage.tsx
+++ b/frontend/src/pages/NotFoundPage.tsx
@@ -8,10 +8,10 @@ export default function NotFoundPage() {
       className="container"
       style={{ textAlign: "center", paddingTop: "4rem", paddingBottom: "4rem" }}
     >
-      <p style={{ fontSize: "4rem", fontWeight: 700, color: "var(--neon-cyan)", margin: 0 }}>
+      <p style={{ fontSize: "var(--text-hero)", fontWeight: 700, color: "var(--neon-cyan)", margin: 0 }}>
         404
       </p>
-      <h1 style={{ fontSize: "1.6rem", margin: "0.75rem 0 0.5rem" }}>Page not found</h1>
+      <h1 style={{ fontSize: "var(--text-heading)", margin: "0.75rem 0 0.5rem" }}>Page not found</h1>
       <p style={{ color: "var(--text-muted)", marginBottom: "2rem" }}>
         The page you&apos;re looking for doesn&apos;t exist.
       </p>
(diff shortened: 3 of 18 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its check_org_membership 502 fix is catalogued in #451. Branch preserved.

PR #435, opened 2026-07-23
refactor: consolidate rate-limit dependencies via factory + fix O(N) purge trigger

lfiaschi:
## What changed

Two-part change from a scheduled deep code-review pass:

**1. DRY-up rate-limit dependencies.** Replaced nine near-identical `_enforce_*_rate_limit` helper functions across `auth_routes.py`, `search_routes.py`, and `registry_routes.py` with a single `rate_limit_dependency(name)` factory in `api/rate_limit.py`. Each dependency still reads `{name}_rate_limit` / `{name}_rate_window` from Settings and caches the limiter on `app.state._{name}_rate_limiter` on first use — runtime behaviour and app.state layout are unchanged. Net −80 lines of copy-pasted route boilerplate. Adding a new rate-limited endpoint is now a one-line `rate_limit_dependency("...")` call, which is what CLAUDE.md already tells contributors to do.

**2. Fix an O(N)-per-request purge trigger in `RateLimiter.__call__`.** The periodic sweep used `sum(len(v) for v in self._requests.values()) % 100`, which is O(N) on every request and non-monotonic — a sweep that shrank the total could then skip subsequent sweeps until the sum grew back. Replaced with a plain `_call_count` counter under the existing lock, and added a regression test that exercises the counter across expired IPs.

Also removed a stray duplicated `# noqa: F401  # noqa: F401` comment in `registry_service.py`.

## Why

This grew out of a scheduled architect/principal-engineer code-review pass over the repo. The rate-limit boilerplate was the highest-leverage, lowest-risk finding — the pattern is called out in CLAUDE.md as required for every new public endpoint, so consolidating it now makes future compliance one line instead of a copy-paste. The purge-trigger fix is a small correctness/perf win in the same file and hot in the request path.

## Other findings from the same review pass (not in this PR)

Filed here for follow-up; each deserves its own focused PR.

**Bugs / correctness**
- `scripts/activate_trackers.py:106-117` — `conn.rollback()` inside a batch loop reverts prior successful inserts in the same transaction; `created` counter overstates.
- `infra/database.py:2043, 2055, 2129` — hybrid FTS/vector search + `fetch_similar_skills` use `ORDER BY <rank> + LIMIT` with no unique tiebreaker; ties are non-deterministic.
- `scripts/backfill_embeddings.py:42-58` — `SELECT ... LIMIT` with no `ORDER BY` (violates the SQL-correctness rule in CLAUDE.md).
- `scripts/backfill_embeddings.py:102` — log message fills both `{}/{}` slots with the same `total_processed` value.
- `domain/publish.py:124-128` — `zf.read(name).decode()` can raise `UnicodeDecodeError`; the publish handler only catches `ValueError`/`BadZipFile`, so non-UTF-8 bytes surface as an HTTP 500.
- `infra/storage.py:172` — `list_eval_log_chunks` doesn't paginate `list_objects_v2`; runs with >1000 chunks silently truncate.

**Missing coverage / DRY**
- `api/search_routes.py:188-204` — `AskSkillRef` is missing `github_forks` even though the pipeline selects it; CLAUDE.md's "keep the ask pipeline in sync with skill metadata" note explicitly lists this file.
- `api/registry_routes.py:959-996` — `get_eval_report_by_skill` and `get_eval_report_by_version_path` are duplicate handlers for the same behaviour.
- `domain/evals.py:100-402` — `run_eval_pipeline` and `stream_eval_pipeline` share ~80% per-case logic (verdict handling, secret redaction, dict construction) — extract a helper.
- No dedicated test files for `api/seo_routes.py`, `api/taxonomy_routes.py`, `domain/publish_pipeline.py`, `domain/repo_utils.py`, or any of the `scripts/*` modules.

**Security / robustness (lower priority)**
- `api/rate_limit.py` keys on `request.client.host`; behind Modal's LB this may collapse all traffic into a single bucket. Any fix here needs care around `X-Forwarded-For` trust — worth a dedicated PR with a documented trusted-proxy policy rather than a drive-by change.
- `api/keys_routes.py:25-29` — `StoreKeyRequest` has no length/format validation on `key_name`/`value`.
- `client/src/dhub/cli/registry.py:468-470` — silent `pass` on non-201/409 tracker POST responses swallows real errors.

**Frontend nits**
- `frontend/src/components/GradeBadge.module.css:13, 19, 25` — hardcoded rem font sizes instead of the `--text-*` scale variables mandated by CLAUDE.md.
- `frontend/src/pages/SkillDetailPage.tsx:195` — `size: content.length` reports character count, not byte size.

## How to test

- `cd server && DHUB_ENV=dev uv run --package decision-hub-server --extra dev pytest tests/ -m "not slow"` — 940 pass locally (233 API + 707 elsewhere), including 12 tests in `tests/test_api/test_rate_limit.py` (7 new for the factory, 2 new for the purge counter).
- `uvx ruff@0.15.3 check .` and `uvx ruff@0.15.3 format --check .` — clean.
- `uvx mypy` on the five touched server modules — clean.

## Checklist
- [x] Tests pass (`make test-server`)
- [x] No breaking API changes (app.state attribute names and per-endpoint behaviour preserved)
- [x] Database migration included (if schema changed) — N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_014pPGvtwKjU32XzYNicNWWa)_

### DIFF
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import rate_limit_dependency
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
@@ -26,16 +26,7 @@ from decision_hub.settings import Settings
 router = APIRouter(prefix="/auth", tags=["auth"])
 
 
-def _enforce_auth_rate_limit(request: Request) -> None:
-    """Rate-limit auth endpoints."""
-    state = request.app.state
-    if not hasattr(state, "_auth_rate_limiter"):
-        settings: Settings = state.settings
-        state._auth_rate_limiter = RateLimiter(
-            max_requests=settings.auth_rate_limit,
-            window_seconds=settings.auth_rate_window,
-        )
-    state._auth_rate_limiter(request)
+_enforce_auth_rate_limit = rate_limit_dependency("auth")
 
 
 # ---------------------------------------------------------------------------
--- a/server/src/decision_hub/api/rate_limit.py
+++ b/server/src/decision_hub/api/rate_limit.py
@@ -3,6 +3,7 @@
 import threading
 import time
 from collections import defaultdict
+from collections.abc import Callable
 
 from fastapi import HTTPException, Request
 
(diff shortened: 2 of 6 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Strict subset of the rate-limiter work in retained #447; its body doubles as the backlog index that #436-#449 implemented. Branch preserved.

PR #436, opened 2026-07-24
fix: harden eval-log storage, search determinism, and manifest I/O

lfiaschi:
## What changed

Six focused, self-contained fixes surfaced by a review pass over the server, shared, and infra layers — each with a covering test.

**Correctness / data-integrity**

- **S3 pagination for eval logs** (`server/src/decision_hub/infra/storage.py`): `list_eval_log_chunks` and `delete_eval_logs` used `list_objects_v2` directly, which caps at 1000 keys per response and never inspects `IsTruncated`. Runs with more than 1000 log chunks silently dropped everything past the first page in the log viewer, and `delete_eval_logs` left those objects behind as orphans. Both now paginate through every page; deletes are batched at the S3-imposed 1000-key `DeleteObjects` limit.
- **Search determinism** (`server/src/decision_hub/infra/database.py`): `search_skills_hybrid` ordered FTS results by `fts_rank DESC` and vector results by `vec_dist ASC` with no unique tiebreaker under `LIMIT`. `ts_rank_cd` ties are common on short queries and cosine distance ties on identical embeddings, so which N rows survived the cut varied non-deterministically across runs and replicas — degrading dedup against the union branch. Same defect in `fetch_similar_skills`. All three ORDER BYs now add `(org_slug, skill_name)` as a stable tiebreaker.
- **`fetch_similar_skills` skipped the archived/removed filter** applied by the main listing, so the "similar skills" sidebar could surface skills whose source repo was deleted or archived on GitHub. Filter now applied consistently.
- **`get_eval_run` race-condition 500** (`server/src/decision_hub/api/registry_routes.py`): the re-read after the zombie check had no `None` guard. A concurrent `versions` delete cascade between the two reads caused `AttributeError` in `_run_to_response(None)` → 500 instead of 404. Guard added.

**Portability**

- **UTF-8 encoding for SKILL.md** (`shared/src/dhub_core/manifest.py`): `parse_skill_md` used `path.read_text()` with no explicit encoding, which uses the platform's locale default (cp1252 on Windows). SKILL.md files with non-ASCII content (accents, emoji, CJK, arrows) mojibaked or raised `UnicodeDecodeError` on Windows. Now reads UTF-8 explicitly.

**Cleanup / DRY**

- Delete unused `AuthToken` dataclass in `server/src/decision_hub/models.py` (grep-verified no callers anywhere in the tree).
- Replace ad-hoc `Fernet(...).decrypt()` call in `run_assessment_background` with the module-level `decrypt_value` helper (which already had tests) — single decrypt path, no behavior change.

## Why

Findings came out of an architect + principal-engineer style review across the four package boundaries (server bugs & correctness, dead-code / DRY, tests coverage, frontend/CLI/shared). This PR picks the highest-leverage, lowest-risk subset — each fix is < 30 lines, in one file, and has a targeted test.

Other findings from the review (rate-limiter keyed on `request.client.host` behind Modal's ingress, orphan-skill row on S3 rollback, tracker visibility filter, CLI subprocess timeouts, frontend SEO base-URL pinned to prod, etc.) are left for follow-ups: they either need product/infra input, span more files than a focused PR warrants, or would benefit from separate discussion.

No issue tracker linked — surfaced by review, not user report.

## How to test

- `make test-shared` — 99 pass, includes new UTF-8 SKILL.md test.
- `make test-server` — 936 pass, includes new S3-pagination and eval-run-vanishes-between-reads tests.
- Manual verification of the pagination fix would need a real eval run with >1000 log chunks; the mock test covers the paginator + batching contract.

## Checklist

- [x] Tests pass (`make test-server`, `make test-shared`, `uvx ruff check .`, `uvx ruff format --check .`)
- [x] No breaking API changes — internal `AuthToken` was unused; all other changes preserve external behavior
- [x] Database migration included (if schema changed) — N/A, no schema changes

---
_Generated by [Claude Code](https://claude.ai/code/session_015vk3Tsj6XMrsZeAYroUq9o)_

### DIFF
--- a/server/src/decision_hub/api/registry_routes.py
+++ b/server/src/decision_hub/api/registry_routes.py
@@ -1132,8 +1132,12 @@ def get_eval_run(
     if run is None or run.user_id != current_user.id:
         raise HTTPException(status_code=404, detail="Eval run not found")
     _check_zombie(conn, run)
-    # Re-read after potential zombie update
+    # Re-read after potential zombie update. A concurrent delete cascade from
+    # ``versions`` can wipe the row between the two reads, so guard against
+    # None here rather than crashing with an AttributeError in _run_to_response.
     run = find_eval_run(conn, parsed_id)
+    if run is None:
+        raise HTTPException(status_code=404, detail="Eval run not found")
     return _run_to_response(run)
 
 
--- a/server/src/decision_hub/api/registry_service.py
+++ b/server/src/decision_hub/api/registry_service.py
@@ -88,8 +88,7 @@ def run_assessment_background(
     log chunks and updates the eval_runs table. Otherwise falls back to
     the original batch pipeline for backward compat.
     """
-    from cryptography.fernet import Fernet
-
+    from decision_hub.domain.crypto import decrypt_value
     from decision_hub.infra.database import create_engine, get_api_keys_for_eval
     from decision_hub.infra.modal_client import get_agent_config, validate_api_key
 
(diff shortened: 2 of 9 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Overlaps the retained merge queue; its UTF-8 manifest read is catalogued in #451. Branch preserved.

PR #429, opened 2026-07-17
Correctness, security, and DRY fixes surfaced by principal-engineer review

lfiaschi:
## What changed

A focused refactor built on a principal-engineer review of the server + CLI. Ships correctness bugs, credential-hygiene fixes, and one big DRY cleanup, each with a locked-in regression test.

**Server**
- **`_check_zombie` no longer kills valid slow runs.** Runs whose heartbeat lags briefly used to flip permanently to `failed` on the next CLI poll (GET endpoints commit their transaction). The read path now only persists `failed` past a 10 min fail threshold; under that, the live status is kept so the worker's next heartbeat resumes normal reporting.
- **`run_streaming_eval` failure-path.** If the pipeline finishes without emitting a `report` event (empty `eval_cases`, aborted generator) the run row no longer hangs in `provisioning`/`running` until the zombie sweep — it's marked `failed` explicitly.
- **`update_org_github_metadata` stops wiping fields.** GitHub returns null for optional org fields (blog, email, description, avatar_url); passing those nulls straight through silently cleared existing values. Now `None` means "skip this column"; empty string is the documented clear-value escape hatch.
- **`list_eval_runs` auth pushed into SQL.** The endpoint used to fetch every row for a `version_id` and filter out foreign users in Python — an info-leak side channel plus unbounded data over the wire. `find_eval_runs_for_version` now takes an optional `user_id=` kwarg wired at the caller.
- **Deterministic `LIMIT` queries.** Added `.id` tiebreakers on the queries flagged by the review — `search_skills_hybrid` (FTS + vector), `fetch_similar_skills`, `find_active_eval_runs_for_user`, `find_eval_runs_for_version`, `list_tracker_metrics`. Satisfies the CLAUDE.md rule.
- **Secret redaction on assessment errors.** `registry_service`'s catch-all now attaches the traceback via `logger.opt(exception=True)` and runs `_redact_secrets` over the message before persisting into `eval_reports` / `eval_runs`, so an API key leaking through an exception message doesn't surface through the eval-run status endpoint.
[...]

PR #439, opened 2026-07-27
fix: crawler S3 orphan, JWT log injection, DRY holistic body reviewers

lfiaschi:
## What changed

Three focused, self-contained fixes surfaced by a fresh deep-review pass over the server. Each ships with a regression test; nothing here changes user-visible behaviour on the healthy path. All 944 server tests + ruff + mypy stay green.

### 1. `crawler/processing.py::_finalize_skill` — orphaned S3 zip on version race (`correctness`)

`_finalize_skill` previously uploaded the zip to S3 first and then called `insert_version`. If a concurrent tracker or crawler container won the version race, `insert_version` raised `IntegrityError`, `VersionConflictError` propagated, and the zip stayed in S3 forever with no DB row pointing at it.

The `execute_publish` path in `publish_pipeline.py` already does the safe ordering (DB row first, S3 last, DB rollback on S3 failure) — the crawler now mirrors it. If S3 upload itself fails, the caller's `conn.commit()` is skipped and the whole transaction rolls back, so neither a stray DB row nor a stray S3 blob is possible.

### 2. `logging._extract_username_from_jwt` — log injection via unverified JWT claim (`security`)

The username claim from the incoming `Authorization: Bearer <jwt>` is read *without verification* (just so request-context log lines can name the caller — real auth happens later in the dependency layer). A hostile or malformed token could carry `\n`, ANSI escapes, or a multi-KB payload that would:

- inject fake log lines (`alice\n2026-01-01 INFO [FAKE ADMIN] "…"`),
- recolor / cursor-move a terminal tailing the JSON logs, or
- bloat every request record with a huge string.

Fix: sanitize the claim to printable ASCII (32–126) and cap at 64 chars before it ever reaches the log context.

### 3. `gemini.review_code_body_safety` / `review_prompt_body_safety` — DRY the retry+parse loop (`code-health`)

Both holistic-review functions had the exact same ~30-line retry / parse / `BodyReviewResult.model_validate` / fail-closed loop copy-pasted. Extracted into a single private `_run_body_review` helper; each caller now just builds its prompt and delegates. Net −60 lines, identical externally-observable behaviour (same fail-closed reasons, same retry cadence).

## Why

No linked issue — findings came from a scheduled review pass. Priors from earlier review PRs (#429–#438) covered rate-limiter, LIKE-escaping, zip hardening, config file perms, etc., so this batch targets what those didn't reach and stays small enough for a single reviewer pass.

## How to test

```bash
uvx ruff@0.15.3 check . && uvx ruff@0.15.3 format --check .   # clean
uvx mypy client/src/ server/src/ shared/src/                    # 88 files, no issues
uv run --package decision-hub-server --extra dev pytest server/tests/ -q -m "not slow" --no-cov   # 944 passed
```

New/updated tests locking each fix in place:

- `server/tests/test_scripts/test_crawler_processing.py::TestPublishOneSkillOrdering` — asserts `upload_skill_zip` is NOT called when `insert_version` raises, and that `insert_version` is called before `upload_skill_zip` on the healthy path.
- `server/tests/test_api/test_logging.py::TestExtractUsernameFromJwt` — 4 new cases: newline stripping, ANSI/control stripping, length cap, non-string claim → empty.
- `server/tests/test_infra/test_gemini.py::TestHolisticBodyReview` — 5 cases across both review entry points: valid response, retry-then-succeed, and both fail-closed paths (unparseable JSON, empty text after retry).

Manual sanity: publishing an unchanged skill continues to short-circuit as `skipped`; a duplicate-version race still surfaces as `VersionConflictError` and the container reports `"skipped"` (unchanged) — just without leaving an S3 object behind.

## Checklist
- [x] Tests pass (`make test`) — 944 server pass, ruff + mypy clean
- [x] No breaking API changes — external shapes unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_01LuvfB9JYsXZhLmBpyDXH9L)_

### DIFF
--- a/server/src/decision_hub/infra/gemini.py
+++ b/server/src/decision_hub/infra/gemini.py
@@ -922,6 +922,64 @@ class BodyReviewResult(BaseModel):
     reason: str
 
 
+def _run_body_review(
+    *,
+    client: dict,
+    model: str,
+    prompt: str,
+    skill_name: str,
+    review_kind: str,
+) -> dict:
+    """Run a holistic Gemini body review with one retry, fail-closed on error.
+
+    Shared engine for ``review_code_body_safety`` and ``review_prompt_body_safety``:
+    both post a single ``BodyReviewResult`` JSON object and treat any error
+    (LLM unreachable, unparseable response, validation failure) as dangerous.
+    ``review_kind`` is a short label (e.g. ``"code"``, ``"body"``) used only
+    in the retry-warning log line.
+    """
+    payload = {
+        "contents": [{"parts": [{"text": prompt}]}],
+        "generationConfig": {"temperature": 0.0},
+    }
+
+    for attempt in range(2):
+        try:
+            data = _gemini_post(client, model, payload)
+
+            text = _extract_text(data)
+            if not text:
+                if attempt == 0:
+                    logger.warning(
+                        "Holistic {} review returned no response for '{}', retrying",
+                        review_kind,
+                        skill_name,
+                    )
+                    continue
+                return {"dangerous": True, "reason": "LLM returned no response (fail-closed)"}
+
+            text = _strip_markdown_fences(text)
+            result = json.loads(text)
+            review = BodyReviewResult.model_validate(result)
+            return review.model_dump()
+        except (json.JSONDecodeError, ValidationError, httpx.HTTPError):
+            if attempt == 0:
+                logger.opt(exception=True).warning(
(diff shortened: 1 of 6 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its crawler S3-orphan ordering and JWT log-injection fixes are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #440, opened 2026-07-28
fix: 5 correctness + hardening fixes from principal-engineer review

lfiaschi:
## What changed

Five focused correctness / hardening fixes surfaced by a fresh full-repo review, all disjoint from what is already covered by open review PRs (#430–#439).

| # | Area | File(s) | Impact |
|---|---|---|---|
| 1 | Tracker preserves `last_error` on partial failure | `domain/tracker_service.py` | Silent-failure fix (**H**) |
| 2 | Ask analytics: DB insert before S3 upload | `api/search_routes.py`, `infra/storage.py` | Orphan-blob prevention (**M**) |
| 3 | Rate limiters on previously-unlimited public GET endpoints | `api/registry_routes.py`, `api/org_routes.py`, `settings.py` | DoS surface (**H**) |
| 4 | `backfill_org_metadata`: token via `GITHUB_TOKEN` env, not argv | `scripts/backfill_org_metadata.py` | Secret hygiene per CLAUDE.md (**M**) |
| 5 | `FileBrowser`: reset expand state on skill navigation | `frontend/src/components/FileBrowser.tsx` | User-visible correctness (**L**) |

### 1. Tracker preserves `last_error` on partial failure

`process_tracker` in `tracker_service.py:804-816` used `all_failed = published_count == 0 and len(errors) > 0`, then cleared `last_error` whenever `not all_failed`. When 3 of 5 skills published and 2 failed, the two failing skills would silently never retry until the repo changed again — and the tracker UI showed no failure. The fix keeps the SHA advance (so successful skills aren't re-published every tick) but records the per-skill errors so operators can see them.

### 2. Ask analytics: DB insert before S3 upload

`_log_ask_analytics` uploaded the log blob to S3 first, then inserted the metadata row. A failed insert (unique-violation, DB down, constraint change) left an orphan blob under `search-logs/` with nothing pointing to it — hard to find, hard to clean up, and it will just accumulate. Extracted `build_search_log_key(log_id)` so the key can be pre-computed and persisted to the DB; the S3 upload runs last so a failed upload leaves a recoverable row (re-uploadable) instead of an unrecoverable blob.

### 3. Rate limiters on previously-unlimited public GET endpoints

Per CLAUDE.md: *"When adding new public endpoints, always add a rate limiter following the existing pattern."* These endpoints slipped through:

- `/v1/stats` (drives `COUNT(DISTINCT …)` + full-scan `DISTINCT` on `skills`)
- `/v1/skills/{org}/{name}/summary`
- `/v1/skills/{org}/{name}/latest-version`
- `/v1/skills/{org}/{name}/eval-report` (both routes)
- `/v1/orgs/stats`
- `/v1/orgs/profiles`
- `/v1/orgs/{slug}/profile`

Introduced a shared `_enforce_public_read_rate_limit` dependency (one per module, both wired to the same settings) backed by two new tunables:

```python
public_read_rate_limit: int = 120   # per-IP burst per window
public_read_rate_window: int = 60   # seconds
```

Ceiling chosen so a normal browser page load (which fans out several summary / latest-version / eval-report requests per skill page) stays comfortably below it while still bounding what a single IP can throw at the DB.

### 4. `backfill_org_metadata`: token via env, not argv

The docstring recommended `--github-token "$(gh auth token)"`, which puts the token on argv — visible to every local user via `ps auxww` and preserved in shell history. Violates CLAUDE.md's *"Sanitize subprocess credentials"* rule. The token is now read from `GITHUB_TOKEN` (or `GH_TOKEN`) first; the legacy `--github-token` flag still works but prints a deprecation warning, and running with no token anywhere exits `2` with a clear message.

### 5. `FileBrowser`: reset expand state on skill navigation

`TreeNodeView`'s `expanded` state is local (`useState`) and the mapping was keyed only on `child.path`. Two skills that share a directory name (like both having `evals/` or `scripts/`) reused each other's node identity and inherited each other's collapse state — so collapsing `evals/` on skill A made it look collapsed on skill B for no reason. Fixed by deriving a `treeKey` from the `files` prop and putting it on the tree container so React fully remounts the tree when the browsed skill changes.

## Why

This is the "principal-engineer review + refactor" scheduled task. I intentionally ran the fresh review against the current `main` and cross-referenced every finding against the ten in-flight `claude/adoring-ride-*` PRs to avoid duplication — the five above are all genuinely new.

## How to test

1. **Tracker** — `uv run --package decision-hub-server --extra dev pytest server/tests/test_domain/test_tracker_service.py -q` (2 partial-failure tests updated to lock in the surfaced-error contract)
2. **Analytics ordering** — `pytest server/tests/test_api/test_search_routes.py::TestLogAskAnalyticsOrdering` (3 tests: happy-path order, insert failure blocks upload, upload failure leaves row with key)
3. **Rate limiters** — `pytest server/tests/test_api/test_public_read_rate_limit.py` (4 tests bursting past the ceiling on each new endpoint)
4. **Storage helper** — `pytest server/tests/test_infra/test_storage_search_log.py` (key format + `upload_search_log`/`build_search_log_key` agreement)
5. **Backfill token** — `pytest server/tests/test_scripts/test_backfill_org_metadata_token.py` (env-var precedence, deprecation warning, hard-fail on missing)
6. **FileBrowser** — `cd frontend && npx vitest run src/components/FileBrowser.test.tsx` (default selection, reset-on-navigation, no expand-state leak between skills)

Full suites:

- `uv run --package decision-hub-server --extra dev pytest server/tests/ -q -m "not slow"` → **947 pass**
- `cd frontend && npx vitest run` → **85 pass** (9 files)
- `uvx ruff@0.15.3 check .` → clean
- `uvx mypy client/src/ server/src/ shared/src/` → `Success: no issues found in 88 source files`

## Checklist
- [x] Tests pass (`make test`) — 947 server + 85 frontend + all lints/mypy
- [x] No breaking API changes — new response fields absent; existing responses unchanged; new settings default to safe values
- [x] Database migration included (if schema changed) — n/a, no schema change

---
_Generated by [Claude Code](https://claude.ai/code/session_01VBv1MuHRj1MYBUp6KniuLH)_

### DIFF
--- a/frontend/src/components/FileBrowser.test.tsx
+++ b/frontend/src/components/FileBrowser.test.tsx
@@ -1,0 +1,68 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import type { SkillFile } from "../types/api";
+import FileBrowser from "./FileBrowser";
+
+const skillA: SkillFile[] = [
+  { path: "SKILL.md", content: "# skill A", size: 10 },
+  { path: "evals/case_a.json", content: "{}", size: 2 },
+  { path: "scripts/run_a.py", content: "print('a')", size: 12 },
+];
+
+const skillB: SkillFile[] = [
+  { path: "SKILL.md", content: "# skill B", size: 10 },
+  { path: "evals/case_b.yaml", content: "case: b", size: 8 },
+  { path: "scripts/run_b.py", content: "print('b')", size: 12 },
+];
+
+describe("FileBrowser", () => {
+  it("renders the SKILL.md content by default", () => {
+    render(<FileBrowser files={skillA} />);
+    expect(screen.getByText(/skill A/)).toBeDefined();
+  });
+
+  it("resets the selected file when navigating to a different skill", () => {
+    // Regression: the FileBrowser's default selection is SKILL.md, so
+    // switching to a new skill should show the new SKILL.md content, not
+    // stale content from the previous skill.
+    const { rerender } = render(<FileBrowser files={skillA} />);
+    expect(screen.getByText(/skill A/)).toBeDefined();
+
+    rerender(<FileBrowser files={skillB} />);
+    expect(screen.getByText(/skill B/)).toBeDefined();
+    expect(screen.queryByText(/skill A/)).toBeNull();
+  });
+
+  it("does not leak expanded-folder state between different skills", () => {
+    // Regression: TreeNodeView holds ``expanded`` in local state and used
+    // to be keyed only on ``child.path``. Two skills that share a directory
+    // name (like ``evals/``) reused the same node identity and inherited
+    // each other's expanded state — meaning a folder collapsed on skill A
+    // would appear collapsed on skill B for no user-visible reason (and
+    // vice versa). Keying the tree container on a per-files identity
+    // forces a full remount so ``expanded`` starts from its default.
+    const { rerender, container } = render(<FileBrowser files={skillA} />);
+
+    // Expand a folder that will collide by path with skill B's evals/.
+    // At depth<2 folders start expanded, so click to COLLAPSE evals in A.
+    const evalsButtonA = screen.getByRole("button", { name: /evals/i });
+    fireEvent.click(evalsButtonA);
+
+    // Sanity: the ``evals/case_a.json`` leaf should no longer be visible.
+    expect(screen.queryByText("case_a.json")).toBeNull();
+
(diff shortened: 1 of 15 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its tracker last_error preservation and ask-analytics DB-before-S3 ordering are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #444, opened 2026-08-01
fix: correctness and safety fixes across server, CLI, and shared

lfiaschi:
## What changed

Nine focused correctness and safety fixes surfaced by a full-codebase review as architect + principal engineer, each with a matching regression test. No behaviour is broadened; no new endpoints or public API surface. Every change is independently reasoned in the diff.

### Server correctness / safety

- **`database._row_to_skill_tracker`** — restore the missing `consecutive_permanent_failures` column. The row mapper silently dropped it, so every tracker returned by `find_skill_tracker` / `claim_due_trackers` / `list_skill_trackers_for_user` carried the dataclass default (`0`), hiding real failure state from every caller that thresholds on it.
- **`database.update_org_github_metadata`** — stop overwriting stored fields with `NULL` when GitHub returns null values (very common for `description` / `blog` / `email`). Only fields the caller explicitly provided reach the `UPDATE`; `github_synced_at` still advances so the sync TTL keeps ticking. An empty string is still treated as an explicit clear.
- **`registry_routes.list_eval_runs`** — filter runs by `user_id` in SQL via the new `find_eval_runs_for_version_and_user` helper instead of pulling every tenant's rows and filtering in Python. Removes cross-tenant work from the DB and cross-tenant data from request memory.
- **`registry_routes.get_eval_run`** — the post-zombie refresh could return `None` on a concurrent delete, crashing `_run_to_response` with a 500. Fall back to the pre-zombie row so the caller still gets a valid response.
- **`tracker_routes.update_tracker`** — replace the production `assert updated is not None` (stripped under `python -O`) with an explicit `HTTPException(404)` so a concurrent delete surfaces as a proper 404 rather than an `AttributeError`.

### CLI robustness

- **`config.get_token` / `get_optional_token`** — strip whitespace from the `DHUB_TOKEN` env var. A trailing newline (common with `export DHUB_TOKEN=$(cat token)`) was ending up inside the `Authorization: Bearer …` header and being rejected by the server with an opaque 400. Whitespace-only values now fall back to the saved config token.
- **`config.save_config`** — chmod the token file to `0600` (POSIX). Previously it inherited the process umask and was often world-readable, exposing the bearer token on shared hosts.
- **`config.extract_error_detail`** — new helper for tolerant error-body parsing. The publish `422` handler now uses it, so a proxy returning `text/plain` no longer crashes the CLI with `JSONDecodeError` and the "Rejected (Grade F): …" message always reaches the user.

### Shared library

- **`manifest.parse_frontmatter_yaml`** — return `{}` instead of `None` for empty or comment-only frontmatter. The caller's `isinstance(data, dict)` check still rejects lists/scalars, and an all-blank frontmatter now surfaces cleanly as `"Required field 'name' is missing"` rather than an `AttributeError` on `.get()`.

## Why

Findings from an architect + principal-engineer review pass across all four workspace packages (`server`, `client`, `shared`, `frontend`) and their test suites. The scope was chosen for the highest correctness / safety leverage per line of change; every fix is small, self-contained, and independently testable. Larger structural findings (rate-limiter proxy-IP handling, eval-log S3 read pattern, publish-pipeline transactional contract, missing `ErrorBoundary`, frontend duplication, test-pyramid gaps) are noted for follow-up PRs rather than mixed in here.

## How to test

- `make test-server` — 940 passed, 34 slow deselected.
- `make test-shared` — 100 passed.
- `make test-client` — 298 passed (excluding `client/tests/test_core/test_docx_integration.py`, whose 5 failures reproduce on the pre-change tree and are unrelated to this refactor).
- `uvx ruff@0.15.3 check .` — all checks passed.
- `uvx ruff@0.15.3 format --check .` — 186 files already formatted.

New regression tests included with each fix:

- `server/tests/test_infra/test_tracker_db.py::TestRowToSkillTracker::test_maps_consecutive_permanent_failures`
- `server/tests/test_infra/test_orgs_db.py` (new file — 4 tests on `update_org_github_metadata`)
- `server/tests/test_api/test_eval_logs.py::TestGetEvalRun::test_get_run_survives_concurrent_delete_after_zombie_check`
- `server/tests/test_api/test_eval_logs.py::TestListEvalRuns::test_list_by_version_scopes_to_current_user_in_sql`
- `server/tests/test_api/test_tracker_routes.py::TestUpdateTracker::test_update_tracker_returns_404_on_concurrent_delete`
- `client/tests/test_cli/test_config.py::TestGetToken::{test_env_var_whitespace_is_stripped,test_whitespace_only_env_var_falls_back_to_config}`
- `client/tests/test_cli/test_config.py::TestSaveConfigPermissions::test_file_has_owner_only_perms`
- `client/tests/test_cli/test_config.py::TestExtractErrorDetail` (4 tests)
- `shared/tests/test_manifest.py::TestParseFrontmatterYaml::{test_empty_frontmatter_returns_empty_dict,test_empty_frontmatter_produces_clean_error_via_parse_skill_md}`

## Checklist

- [x] Tests pass (`make test` on affected suites; server + shared + non-docx CLI green)
- [x] No breaking API changes — response shapes and status codes unchanged on happy paths; only failure modes improve (500→404, 500→200-with-valid-body, silent NULL-clobber→no-op)
- [x] No database migration needed — no schema changes

---
_Generated by [Claude Code](https://claude.ai/code/session_019Xvpjdyw5mQvXhQgqbHv1P)_

### DIFF
--- a/client/src/dhub/cli/config.py
+++ b/client/src/dhub/cli/config.py
@@ -1,5 +1,6 @@
 """CLI configuration file management for ~/.dhub/config.{env}.json."""
 
+import contextlib
 import json
 import os
 from dataclasses import asdict, dataclass
@@ -82,7 +83,9 @@ def load_config() -> CliConfig:
 def save_config(config: CliConfig) -> None:
     """Save CLI config to ~/.dhub/config.{env}.json.
 
-    Creates the ~/.dhub directory if it does not already exist.
+    Creates the ~/.dhub directory if it does not already exist and writes
+    the file with 0600 permissions (owner read/write only) so the bearer
+    token is not world-readable on shared machines.
     """
     CONFIG_DIR.mkdir(parents=True, exist_ok=True)
     path = config_file()
@@ -90,6 +93,11 @@ def save_config(config: CliConfig) -> None:
         json.dumps(asdict(config), indent=2) + "\n",
         encoding="utf-8",
     )
+    # POSIX-only. Windows has no meaningful equivalent for a config file
+    # in a per-user home directory, so silently skip if chmod is a no-op
+    # or errors.
+    with contextlib.suppress(OSError, NotImplementedError):
+        path.chmod(0o600)
 
 
 def get_api_url() -> str:
@@ -103,12 +111,19 @@ def get_api_url() -> str:
 def get_token() -> str:
     """Get the auth token from DHUB_TOKEN env var or saved config.
 
+    Env-var values are stripped so a stray newline (common with
+    ``export DHUB_TOKEN=$(cat token)``) does not become part of the
+    ``Authorization: Bearer …`` header — httpx would emit an invalid
+    request and the server would reject it with an opaque 400.
+
     Raises:
         typer.Exit: If no token is available (user not logged in).
     """
     env_token = os.environ.get("DHUB_TOKEN")
     if env_token:
-        return env_token
+        stripped = env_token.strip()
+        if stripped:
+            return stripped
 
     token = load_config().token
     if not token:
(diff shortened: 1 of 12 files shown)


lfiaschi:
Closing as part of the 2026-08-11 open-PR consolidation (see #451). Its org-metadata NULL-clobber, DHUB_TOKEN newline, and cross-tenant SQL fixes are catalogued in #451 for the follow-up best-of PR. Branch preserved.

PR #447, opened 2026-08-04
Fix rate-limiter races, retry bugs, and other correctness issues

lfiaschi:
## What changed

Fixes surfaced by a full-codebase review across architecture, bugs, security, performance, testing, and code health. Ten concrete defects plus a handful of dead-code and DRY cleanups — all self-contained, no schema changes, no API-breaking rename.

## Why

Not tied to a single issue. Highlights (each cited with `file:line` in the commit message):

**Rate limiter** — `api/rate_limit.py`
- **Race condition on lazy init.** Eight route handlers used `if not hasattr(state, "_x_limiter"): state._x_limiter = RateLimiter(...)`. FastAPI runs sync deps in a threadpool, so two concurrent first-hits both saw `hasattr == False`, both built a `RateLimiter`, and one silently overwrote the other — dropping the loser's counter and letting a burst through. Replaced with a `RateLimiters` container built eagerly in `create_app()` from settings, and a `limiter_dep("name")` helper that looks limiters up per-request.
- **O(n_ips) work per request under the lock.** `_purge_stale` was gated on `sum(len(v) for v in self._requests.values()) % 100 == 0`, which (a) did O(n_ips) work under the lock every request, and (b) could easily never hit 100 exactly (so the purge fired at unpredictable intervals). Replaced with a monotonic `_request_count` incremented on admitted requests only.
- **Boilerplate.** Removed eight nearly identical `_enforce_*_rate_limit` functions. Each route is now one `Depends(limiter_dep("name"))` line.

**Missing rate limits on public endpoints (CLAUDE.md rule)**
Added limiters to six previously-unbounded GET endpoints: `/v1/stats`, `/v1/skills/{}/{}/summary`, `/v1/skills/{}/{}/latest-version`, `/v1/skills/{}/{}/eval-report`, and the versioned eval-report path. Each did 2–3 DB round-trips per hit with no per-IP cap.

**Correctness bugs**
- **Gemini retried permission-denied 403s.** The retry list treated every 403 as transient, so requests with a rotated-but-not-updated API key wasted ~7 s before failing. 403 is now only retried when the response body contains rate/quota indicators.
- **GitHub `graphql()` bypassed the pooled client.** It called module-level `httpx.post` instead of `self._client.post`, forcing a fresh TCP+TLS handshake per batch.
- **`backfill_embeddings.py` had `LIMIT` without `ORDER BY`.** Violated the CLAUDE.md SQL rule; on retry after a partial commit, PostgreSQL could pick a different row set and hot-loop. Added `ORDER BY skills.id`.
- **Crawler checkpoint non-atomic save.** `path.write_text(...)` in-place could leave a truncated JSON on SIGKILL/OOM despite the "crash-safe" docstring. Now writes to `<path>.tmp` then `os.replace`s into place.

**Small perf / cleanups**
- Anthropic judge regex compiled per call — moved to module scope.
- Dead code removed: `domain/orgs.py::validate_role`, `TTLCache.invalidate/clear` (all had zero application callers, verified by grep).
- DRY: extracted duplicated `_parse_uuid` into `api/deps.py::parse_uuid_or_422`.

## How to test

```bash
make test        # 933 server + 98 shared + client suites pass
make lint        # ruff check + format --check + tsc + eslint
make typecheck   # mypy across 88 source files
```

Concrete regressions to look at:
- `server/tests/test_api/test_rate_limit.py` — `test_purge_uses_monotonic_counter`, `test_over_limit_admissions_do_not_advance_counter`, and the new `TestRateLimiters` / `TestLimiterDep` classes.
- `server/tests/test_infra/test_gemini.py::TestGeminiPostRetry::test_permission_denied_403_fails_fast`.
- `server/tests/test_infra/test_github_client.py::TestGitHubClientGraphQL` — asserts the pooled client is used.
- `server/tests/test_scripts/test_crawler_checkpoint.py` — new file, atomic-write behaviour + prior-checkpoint survival on simulated crash.
- `server/tests/test_scripts/test_backfill_embeddings.py` — new file, asserts `ORDER BY ... LIMIT` in the batch SELECT.

## Checklist
- [x] No breaking API changes (route paths unchanged, response shapes unchanged, added rate limits use existing settings)
- [x] Database migration included (if schema changed) — N/A, no schema changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01UsgHnqavEHM47e2EjfoWm1)_

### DIFF
--- a/server/src/decision_hub/api/app.py
+++ b/server/src/decision_hub/api/app.py
@@ -12,6 +12,7 @@ from loguru import logger
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from decision_hub.api.deps import get_current_user
+from decision_hub.api.rate_limit import RateLimiters
 from decision_hub.infra.database import create_engine
 from decision_hub.infra.storage import create_s3_client
 from decision_hub.logging import RequestLoggingMiddleware, setup_logging
@@ -158,6 +159,10 @@ def create_app() -> FastAPI:
     app.state.engine = engine
     app.state.settings = settings
     app.state.s3_client = s3_client
+
+    # Rate limiters are constructed eagerly here so route dependencies can
+    # look them up by name without racing (see api/rate_limit.py).
+    app.state.rate_limiters = RateLimiters(settings)
 
     # In-memory TTL cache for hot read paths (per-container, not shared
     # across Modal replicas).
--- a/server/src/decision_hub/api/auth_routes.py
+++ b/server/src/decision_hub/api/auth_routes.py
@@ -1,13 +1,13 @@
 """Authentication routes - GitHub Device Flow login."""
 
 import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException
 from loguru import logger
 from pydantic import BaseModel
 from sqlalchemy.engine import Connection
 
 from decision_hub.api.deps import get_connection, get_engine, get_settings
-from decision_hub.api.rate_limit import RateLimiter
+from decision_hub.api.rate_limit import limiter_dep
 from decision_hub.domain.auth import create_jwt
 from decision_hub.domain.orgs import sync_org_github_metadata, sync_user_orgs
 from decision_hub.infra.database import upsert_user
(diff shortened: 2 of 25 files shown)


lfiaschi:
Pushed 751d18c fixing `_is_retriable_403` in `infra/gemini.py`.

The check scanned the uppercased body for the bare substring `"RATE"`, which also matches `"GENERATE"` and `"SEPARATE"` — so a permanent `PERMISSION_DENIED` 403 whose message mentions generating content still burned the full retry budget, the exact latency this check was added to remove. It now matches Google's canonical `error.status` (`RESOURCE_EXHAUSTED`, `UNAVAILABLE`) first and falls back to whole phrases in `error.message` for quota errors with no machine-readable status. Two regression tests added; server suite passes (935 passed, 34 skipped).

Separately, I checked the `graphql()` change that swapped the manual `Authorization` header for `self._client.post`: `_client` is built with `Authorization: Bearer {token}` in its default headers (`github_client.py:33-41`), so auth is preserved and the per-request `Accept: application/json` correctly overrides only the Accept default. No change needed.

